PR #4, opened 2022-04-08
App service implementation

wira-mms:
- [x] GetAllActiveApps()
- [x] GetFavouritesForCurrentUser()
- [x] SetFavouriteForCurrentUser()
- [x] RemoveFavouriteForCurrentUser()
- [x] Add authorization
- [x] Documentation
- [ ] Tests


src/marketplace/CatenaX.NetworkServices.App.Service/ViewModels/AppViewModel.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+        public Guid Id { get; set; }
+
+        public string Title { get; set; } = default!;
+
+        public string ShortDescription { get; set; } = default!;
+
+        public string Provider { get; set; } = default!;
+
+        public ICollection<string> UseCases { get; set; } = new HashSet<string>();
```
ntruchsess, line 13: I suggest the use or IEnumerable with default 'Enumerable.Empty' instead of ICollection to indicate it is not intended to add anything to the property but set (and thus replace) the reference entirely.

ntruchsess (changes requested):
please see me comments and check https://github.com/catenax-ng/product-portal-backend/pull/10 for a more efficient version

src/marketplace/CatenaX.NetworkServices.App.Service/BusinessLogic/AppsBusinessLogic.cs
```diff
@@ -0,0 +1,106 @@
...
+        public async Task<IEnumerable<Guid>> GetAllFavouriteAppsForUserAsync(Guid userId)
+        {
+            return await this.context.IamUsers.AsNoTracking()
+                .Include(u => u.CompanyUser!.Apps)
+                .Where(u => u.UserEntityId == userId.ToString()) // Id is unique, so single user
+                .SelectMany(u => u.CompanyUser!.Apps.Select(a => a.Id))
+                .ToListAsync();
+        }
+
+        public async Task RemoveFavouriteAppForUserAsync(Guid appId, Guid userId)
```
ntruchsess, line 59: please use string as type of userId. While the userIds we retrive from Keycloak by default look like guids there is no guarantee it will always be a guid. The field in table 'user_entity' (see the schema of keycloak-db) is of type varchar(36) and you may pass in whatever 36char string on creation of a new keycloak user. It just happens that keycloak makes use of guids if you do not pass in your id of choice.

src/marketplace/CatenaX.NetworkServices.App.Service/BusinessLogic/AppsBusinessLogic.cs
```diff
@@ -0,0 +1,106 @@
...
+            .Where(u => u.UserEntityId == userId.ToString())
+            .Select(u => u.CompanyUser)
+            .SingleOrDefaultAsync();
+
+            if(companyUser is null)
+            {
+                throw new ArgumentOutOfRangeException(nameof(userId));
+            }
+
+            var app = companyUser.Apps.FirstOrDefault(a => a.Id == appId);
```
ntruchsess, line 72: why not just remove the entry from CompanyUserAppAssignedFavorites like 'context.Delete(new CompanyUserAppAssignedFavorite { CompanyUserId = companyUserId, AppId = appId };'
There really is no benefit in retrieving any of the linked entities before. If you really(!) want to know the entry actually exists before deleting it you'd use context.CompanyUserAppAssignedFavorites.FindAsync( new CompanyUserAppAssignedFavorite { .... }) before and delete the retrieved entity but this will only work reliably when being enclosed in a single transaction-context with the deletion. (The entry might be removed from the db in a parallel thread while this one executes). As the outcome is correct even if the entry is not existing before the deletion (and we certainly do not want to lock that on the database if it really does not make a difference) there is no point in first reading then deleting an entry that should be gone after the call anyway.

BTW: I'm pretty sure that can be done in a single call to the database combining the retrieval of companyuser with the deletion from CompanyUserAppAssingedFavorites.  @jjeroch:  It would be even better if we store the company-user as an Attribute in Keycloak and map it into the token so any application might access that value without going to the database.

src/marketplace/CatenaX.NetworkServices.App.Service/Controllers/AppsController.cs
```diff
@@ -0,0 +1,73 @@
...
+        {
+            this.logger = logger;
+            this.appsBusinessLogic = appsBusinessLogic;
+        }
+
+        [HttpGet]
+        [Route("active")]
+        public async Task<ActionResult<IEnumerable<AppViewModel>>> GetAllActiveApps([FromQuery] string? lang = null)
+        {
+            return Ok(await this.appsBusinessLogic.GetAllActiveAppsAsync(lang).ToListAsync().ConfigureAwait(false));
```
ntruchsess, line 23: I checked the dotnet core 6 documentation. You may just return the IAsyncEnumerable as it is (no await, no ToListAsync) - the framework will make sure it is being streamed to the client

src/marketplace/CatenaX.NetworkServices.App.Service/Controllers/AppsController.cs
```diff
@@ -0,0 +1,73 @@
...
+        }
+
+        [HttpGet]
+        [Route("favourites/for/user/{userId}")]
+        public async Task<ActionResult<IEnumerable<Guid>>> GetAllFavouriteAppsForUser([FromRoute] Guid userId)
```
ntruchsess, line 28: same as above: please use 'string' as type of parameter 'userId' for the reasons given above

ntruchsess (changes requested):
see my comments inline

src/marketplace/CatenaX.NetworkServices.App.Service/BusinessLogic/AppsBusinessLogic.cs
jjeroch: @ntruchsess with your statement above 
`_@jjeroch: It would be even better if we store the company-user as an Attribute in Keycloak and map it into the token so any application might access that value without going to the database._`
I am guessing you are talking about the company_user_id given by us, right? Good idea. I am running a quick check later how many services this will impact. I believe its a small number and we can discuss jointly when such an improvement of the logic would make sense. Likely with the next "company user id" ticket

src/marketplace/CatenaX.NetworkServices.App.Service/Controllers/AppsController.cs
```diff
@@ -1,73 +1,150 @@
...
+                return StatusCode((int)HttpStatusCode.InternalServerError);
             }
 
             return Ok();
         }
+
+        private Guid GetIamUserIdFromClaims()
+        {
+            var subClaimValue = User.Claims.SingleOrDefault(c => c.Type == "sub")?.Value;
+            _ = Guid.TryParse(subClaimValue, out var userId);
```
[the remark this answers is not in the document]
wira-mms: https://github.com/catenax-ng/product-portal-backend/pull/4/commits/1f99a887dce0ac9cb65c37f9e5a6210574aca0d4

src/marketplace/CatenaX.NetworkServices.App.Service/BusinessLogic/AppsBusinessLogic.cs
wira-mms: https://github.com/catenax-ng/product-portal-backend/pull/4/commits/1f99a887dce0ac9cb65c37f9e5a6210574aca0d4

src/marketplace/CatenaX.NetworkServices.App.Service/Controllers/AppsController.cs
wira-mms: https://github.com/catenax-ng/product-portal-backend/pull/4/commits/afb9ebc010b9838d6366da826d8cbd25a770dde0
Although I am unsure whether this is a good idea because it's not obviously async anymore.

src/marketplace/CatenaX.NetworkServices.App.Service/BusinessLogic/IAppsBusinessLogic.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// <param name="languageShortName">Optional two character language specifier for the app description. No description if not provided.</param>
+        /// <returns>List of active marketplace apps.</returns>
+        public IAsyncEnumerable<AppViewModel> GetAllActiveAppsAsync(string? languageShortName = null);
+
+        /// <summary>
+        /// Get IDs of all favourite apps of the user by ID.
+        /// </summary>
+        /// <param name="userId">ID of the user to get favourite apps for.</param>
+        /// <returns>List of IDs of user's favourite apps.</returns>
+        public Task<IEnumerable<Guid>> GetAllFavouriteAppsForUserAsync(string userId);
```
ntruchsess, line 22: return type should be `IAsyncEnumerable<Guid>`

src/marketplace/CatenaX.NetworkServices.App.Service/Controllers/AppsController.cs
```diff
@@ -0,0 +1,146 @@
...
+        /// </summary>
+        /// <param name="lang" example="en">Optional two character language specifier for the app description. Will be empty if not provided.</param>
+        /// <returns>Collection of all active marketplace apps.</returns>
+        /// <remarks>Example: GET: /api/apps/active</remarks>
+        /// <response code="200">Returns the list of all active marketplace apps.</response>
+        [HttpGet]
+        [Route("active")]
+        [Authorize(Roles = "view_apps")]
+        [ProducesResponseType(typeof(IEnumerable<AppViewModel>), StatusCodes.Status200OK)]
+        public ActionResult<IEnumerable<AppViewModel>> GetAllActiveApps([FromQuery] string? lang = null)
```
ntruchsess, line 42: return type should be `IAsyncEnumerable<AppViewModel>`, return-type ActionResult seems no longer be required (see e.g. https://www.tpeczek.com/2021/07/aspnet-core-6-and-iasyncenumerable.html )

src/marketplace/CatenaX.NetworkServices.App.Service/Controllers/AppsController.cs
```diff
@@ -0,0 +1,146 @@
...
+        /// <returns>Collection of IDs of favourite apps.</returns>
+        /// <remarks>Example: GET: /api/apps/favourites</remarks>
+        /// <response code="200">Returns the list of favourite apps of current user.</response>
+        /// <response code="400">If sub claim is empty/invalid.</response>
+        [HttpGet]
+        [Route("favourites")]
+        [Authorize(Roles = "view_apps")]
+        [ProducesResponseType(typeof(IEnumerable<Guid>), StatusCodes.Status200OK)]
+        [ProducesResponseType(typeof(string), StatusCodes.Status400BadRequest)]
+        public async Task<ActionResult<IEnumerable<Guid>>> GetAllFavouriteAppsForCurrentUserAsync()
```
ntruchsess, line 59: return type should be `IAsyncEnumerable<Guid>`, return-type ActionResult seems no longer be required (see e.g. https://www.tpeczek.com/2021/07/aspnet-core-6-and-iasyncenumerable.html )

ntruchsess (changes requested):

src/marketplace/CatenaX.NetworkServices.App.Service/BusinessLogic/IAppsBusinessLogic.cs
wira-mms: https://github.com/catenax-ng/product-portal-backend/pull/4/commits/a4783f2608c8c19704456a3f4e8e5d17400bda62

src/marketplace/CatenaX.NetworkServices.App.Service/Controllers/AppsController.cs
wira-mms: https://github.com/catenax-ng/product-portal-backend/pull/4/commits/a4783f2608c8c19704456a3f4e8e5d17400bda62

ntruchsess (approved):

PR #8, opened 2022-04-08
CPLP-362: refactor bpn-endpoints in useradmin-service to efcore

ntruchsess:


ntruchsess:
@evegufy : I managed to test the modified endpoints successfully. There were some issues with the translation of Linq to SQL where the result was semantically a bit different to what I expected. There also was an issue with running a query within an asynchronous iteration of the previous result that let to an error as you cannot execute a query in eccore on the same instance of DBContext while another query is still open.
I did enable the sql-logging in appsettings.Development.json - that is quite interesting to see what EFCore translates Linq into.  configfiles for devenv/useradminservice in Keepass are updated accordingly.
From my point of view this PR is now ready to merge.

evegufy (approved):

PR #9, opened 2022-04-08
CPLP-362: update and retrieve application status

ntruchsess:
This PR implements:
https://jira.catena-x.net/browse/CPLP-621
https://jira.catena-x.net/browse/CPLP-752


evegufy:
Retested both endpoints (PUT and GET) successfully on the local environment.
Note: the id column in the table company_applications is still called 'id' not 'application_id', also in the currently single source of truth: https://github.com/tractusx-team-portal-onboarding/devenv/blob/master/postgres/init/20-portal.sql#L376

evegufy:
Retested both endpoints (PUT and GET) successfully on the local environment.
Note: the id column in the table company_applications is still called 'id' not 'application_id', also in the currently single source of truth: https://github.com/tractusx-team-portal-onboarding/devenv/blob/master/postgres/init/20-portal.sql#L376

evegufy (approved):

PR #10, opened 2022-04-09
CPLP-679: more efficient linq limits the amount of data to what is actually needed

ntruchsess:
the query results generation of in the following sql:

      SELECT a2.id, a2.name, c.name, a2.thumbnail_url, (
          SELECT a.description_short
          FROM portal.app_descriptions AS a
          WHERE (a2.id = a.app_id) AND (a.language_short_name = @__languageShortName_0)
          LIMIT 1), (
          SELECT a1.licensetext
          FROM portal.app_assigned_licenses AS a0
          INNER JOIN portal.app_licenses AS a1 ON a0.app_license_id = a1.id
          WHERE a2.id = a0.app_id
          LIMIT 1), c.id, t.name, t.app_id, t.use_case_id, t.id
      FROM portal.apps AS a2
      LEFT JOIN portal.companies AS c ON a2.vendor_company_id = c.id
      LEFT JOIN (
          SELECT u.name, a3.app_id, a3.use_case_id, u.id
          FROM portal.app_assigned_use_cases AS a3
          INNER JOIN portal.use_cases AS u ON a3.use_case_id = u.id
      ) AS t ON a2.id = t.app_id
      WHERE (a2.date_released IS NOT NULL) AND (a2.date_released <= now() AT TIME ZONE 'UTC')
      ORDER BY a2.id, c.id, t.app_id, t.use_case_id, t.id

setting 'languageShortName' to null dynamically eliminates the part that selects description_short and the generated sql is a bit simpler:

      SELECT a1.id, a1.name, c.name, a1.thumbnail_url, (
          SELECT a0.licensetext
          FROM portal.app_assigned_licenses AS a
          INNER JOIN portal.app_licenses AS a0 ON a.app_license_id = a0.id
          WHERE a1.id = a.app_id
          LIMIT 1), c.id, t.name, t.app_id, t.use_case_id, t.id
      FROM portal.apps AS a1
      LEFT JOIN portal.companies AS c ON a1.vendor_company_id = c.id
      LEFT JOIN (
          SELECT u.name, a2.app_id, a2.use_case_id, u.id
          FROM portal.app_assigned_use_cases AS a2
          INNER JOIN portal.use_cases AS u ON a2.use_case_id = u.id
      ) AS t ON a1.id = t.app_id
      WHERE (a1.date_released IS NOT NULL) AND (a1.date_released <= now() AT TIME ZONE 'UTC')
      ORDER BY a1.id, c.id, t.app_id, t.use_case_id, t.id

besides a few primary-keys (c.id, t.app_id, t.use_case_id, t.id which are not referenced in the query but added to the select by ef-core (for reasons that I do not yet fully understand) this sql does not return any redundant data

### DIFF
--- a/src/marketplace/CatenaX.NetworkServices.App.Service/BusinessLogic/AppsBusinessLogic.cs
+++ b/src/marketplace/CatenaX.NetworkServices.App.Service/BusinessLogic/AppsBusinessLogic.cs
@@ -13,29 +13,37 @@ namespace CatenaX.NetworkServices.App.Service.BusinessLogic
             this.context = context;
         }
 
-        public async Task<IEnumerable<AppViewModel>> GetAllActiveAppsAsync(string? languageShortName = null)
+        public async IAsyncEnumerable<AppViewModel> GetAllActiveAppsAsync(string? languageShortName = null)
         {
-            var activeApps = await this.context.Apps.AsNoTracking()
-                .Include(a => a.AppDescriptions)
-                .Include(a => a.VendorCompany)
-                .Include(a => a.UseCases)
-                .Include(a => a.AppLicenses)
-                .Where(a => a.DateReleased.HasValue && a.DateReleased <= DateTime.UtcNow)
-                .ToListAsync();
-
-            var mappedApps = activeApps.Select(a => new AppViewModel
-            {
-                Id = a.Id,
-                Title = a.Name ?? "",
-                Provider = a.VendorCompany?.Name ?? "",
-                UseCases = a.UseCases.Select(uc => uc.Name ?? "").ToList(),
-                LeadPictureUri = a.ThumbnailUrl ?? "",
-                ShortDescription = languageShortName is null ? 
-                    "" : 
-                    a.AppDescriptions.FirstOrDefault(ad => ad.AppId == a.Id && ad.LanguageShortName == languageShortName)?.DescriptionShort ?? "",
-                Price = a.AppLicenses.FirstOrDefault()?.Licensetext ?? ""
-            });
-            return mappedApps;
+            await foreach(var app in context.Apps.AsQueryable()
+                .Where(app => app.DateReleased.HasValue && app.DateReleased <= DateTime.UtcNow)
+                .Select(a => new {
+                    Id = a.Id,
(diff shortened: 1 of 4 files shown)


src/marketplace/CatenaX.NetworkServices.App.Service/Controllers/AppsController.cs
```diff
@@ -20,7 +20,7 @@ public AppsController(ILogger<AppsController> logger, IAppsBusinessLogic appsBus
         [Route("active")]
         public async Task<IActionResult> GetAllActiveApps([FromQuery] string? lang = null)
         {
-            return Ok(await this.appsBusinessLogic.GetAllActiveAppsAsync(lang));
+            return Ok(await this.appsBusinessLogic.GetAllActiveAppsAsync(lang).ToListAsync().ConfigureAwait(false));
```
ntruchsess, line 23: actually I think there should be a way to stream the IAsyncEnumerable into the result instead of converting it into a list. This would allow to very efficiently return huge numbers of entries it would stream directly from the dabase-cursor instead of buffering the whole result in memory first.

src/marketplace/CatenaX.NetworkServices.App.Service/BusinessLogic/AppsBusinessLogic.cs
```diff
@@ -13,29 +13,38 @@ public AppsBusinessLogic(PortalDBContext context)
...
+                    LicenseText = a.AppLicenses
+                        .Select(license => license.Licensetext)
+                        .FirstOrDefault()
+                }).AsAsyncEnumerable())
+                {
+                    yield return new AppViewModel {
+                        Id = app.Id,
+                        Title = app.Name ?? "",
+                        Provider = app.VendorCompanyName ?? "",
+                        UseCases = app.UseCaseNames.Select(name => name ?? "").ToList(),
```
ntruchsess, line 42: if UseCases would be changed to type IEnumerable<string> calling to 'ToList()' would not be required - the IEnumerable would be iterated when being serialized into json.

ntruchsess (commented):
just left a few comments on my own code suggesting further changes that were not in scope of this change.

PR #12, opened 2022-04-12
Update nextsteps.html

jjeroch:
@Akshay-Chaudhari-BMW  I have updated the spaces between the header and the body. However I feel that the email layout might differ in your outlook compared to mine. Could you generate the "Next Steps" Email for my Email Address? I like to see how the email actually looks. Which Email Client are you using? Is it a old Outlook Client?


Akshay-Chaudhari-BMW:
Email template format may vary by outlook versions.

PR #15, opened 2022-04-13
CPLP-613: set userEntityId to string to match keycloaks internal datamodel

ntruchsess:


ntruchsess:
as requested changes have been manually implemented in target branch this PR is obsolete. Will discuss not implemented rename of column 'user_entity_id' to 'shared_user_entity_id' (that I did apply to point out the referenced userid does not refer to central keycloak but shared keycloak) after my return.

PR #23, opened 2022-04-26
CPLP-627 endpoint returning consentStatus and agreementType

ntruchsess:
implements https://jira.catena-x.net/browse/CPLP-627


ntruchsess:
@jjeroch : I requested your review as I'm not sure what the field 'applicationType' which is requested by the userstory does actually stand for as it is null in all existing testdata-sets. (Technically this is not an issue - if being null it is not included in the response, but maybe this is not what we actually do need here). I think we need to clarify this detail before merging.

jjeroch:
@ntruchsess 'agreement_type' right?
Actually its the consent.agreement_id which is connected to the agreements.agreement_type. We need the agreements.agreement_type , to know to which agreement the company gave their consent

evegufy (approved):
Approved from a technical POV

ntruchsess:
> @ntruchsess 'agreement_type' right? Actually its the consent.agreement_id which is connected to the agreements.agreement_type. We need the agreements.agreement_type , to know to which agreement the company gave their consent
@jjeroch : yes, agreement.agreement_type (that was a type in my initial question). I'm asking that question as I do not know what are the allowed values of agreement_type as I want to update the field when the consent is given (to implement https://jira.catena-x.net/browse/CPLP-626). Is it possible that this field should have semantics of an enumeration and we just forgot to convert this to an agreement_type_id plus separate table agreement_types (columns agreement_type_id and label)?

ntruchsess:
closing, merged into https://github.com/catenax-ng/product-portal-backend/pull/27

PR #30, opened 2022-05-04
Feature/CPLP- 625 retrieve invited users

qxz2mqe:
CPLP- 625 retrieve invited users


src/portalbackend/CatenaX.NetworkServices.PortalBackend.DBAccess/PortalBackendDBAccess.cs
```diff
@@ -206,5 +210,22 @@ public Task<CompanyApplicationStatusId> GetApplicationStatusAsync(Guid applicati
...
         public Task<int> SaveAsync() =>
             _dbContext.SaveChangesAsync();
+
+        public IAsyncEnumerable<InvitedUsers> GetInvitedUsersDetail(Guid applicationId)
+        {
+            var result = (from invitation in _dbContext.Invitations
+                          join invitationStatus in _dbContext.InvitationStatuses on invitation.InvitationStatusId equals invitationStatus.InvitationStatusId
+                          join companyuser in _dbContext.CompanyUsers on invitation.CompanyUserId equals companyuser.Id
+                          join iamuser in _dbContext.IamUsers on companyuser.Id equals iamuser.CompanyUserId
+                          where invitation.CompanyApplicationId.ToString() == applicationId.ToString()
```
ntruchsess, line 220: those calls of ToString seem to be redundant. Both sides are Guid (also on the database) and should be compared without any conversion

src/portalbackend/CatenaX.NetworkServices.PortalBackend.DBAccess/Models/InvitedUsers.cs
```diff
@@ -0,0 +1,13 @@
+using System;
+using System.Collections.Generic;
+
+namespace CatenaX.NetworkServices.PortalBackend.DBAccess.Models
+{
+    public class InvitedUsers
```
ntruchsess, line 6: I suggest renaming to 'InvitedUser' (without the trailing 's'). I refers to a single instance

src/portalbackend/CatenaX.NetworkServices.PortalBackend.DBAccess/Models/InvitedUsers.cs
```diff
@@ -0,0 +1,13 @@
...
+    {
+        public string EmailId { get; set; }
+        public string InvitationStatus { get; set; }
+        public string UserId { get; set; }
```
ntruchsess, line 10: I'm not sure whether including the UserId is required or even helpful. (Of course it will do no harm ;-))
It's not mentioned in the userstory, please clarify with @jjeroch

src/portalbackend/CatenaX.NetworkServices.PortalBackend.DBAccess/Models/InvitedUsers.cs
```diff
@@ -0,0 +1,13 @@
...
+        public IEnumerable<Roles> InvitedUserRoles { get; set; }
```
ntruchsess, line 11: I suggest to return the Role-names only:
`public IEnumerable<string> InvitedUserRoles { get; set; }`
The userstory doesn't mention the roles id and the boolean clientrole should be part of the response.
Please also clarify with @jjeroch

src/registration/CatenaX.NetworkServices.Registration.Service/BusinessLogic/IRegistrationBusinessLogic.cs
```diff
@@ -33,5 +33,6 @@ public interface IRegistrationBusinessLogic
         Task<int> SetApplicationStatusAsync(Guid? applicationId, CompanyApplicationStatusId? status);
         Task<CompanyApplicationStatusId> GetApplicationStatusAsync(Guid? applicationId);
         Task<bool> SubmitRegistrationAsync(string userEmail);
+        Task<InvitedUserRoleMapper> GetInvitedUsersDetail(Guid applicationId);
```
ntruchsess, line 36: That should rather be:
`IAsyncEnumerable<InvitedUser> GetInvitedUsersDetail(Guid applicationId);`
InvitedUserRoleMapper is just an wrapper around the IEnumerable<InvitedUsers> plus instead of returning Task<IEnumerable<...>> it's always more efficient to return IAsyncEnumerable<...> (of course only if it is feasable)

src/registration/CatenaX.NetworkServices.Registration.Service/BusinessLogic/RegistrationBusinessLogic.cs
```diff
@@ -342,8 +344,40 @@ public async Task<bool> SubmitRegistrationAsync(string userEmail)
                 { "url", $"{_settings.BasePortalAddress}"},
             };
 
-            await _mailingService.SendMails(userEmail,mailParameters, new List<string> { "SubmitRegistrationTemplate" });
+            await _mailingService.SendMails(userEmail, mailParameters, new List<string> { "SubmitRegistrationTemplate" });
             return true;
         }
+
+        public async Task<InvitedUserRoleMapper> GetInvitedUsersDetail(Guid applicationId)
```
ntruchsess, line 351: I suggest:
```
        public async IAsyncEnumerable<InvitedUsers> GetInvitedUsersDetail(Guid applicationId)
        {
            await foreach (var item in _portalDBAccess.GetInvitedUsersDetail(applicationId).ConfigureAwait(false))
            {
                var response = await _provisioningManager.GetClientRoleMappingsForUserAsync(item.UserId, _settings.KeyCloakClientID);
                var invitedUserRoles = response.Select(index =>
                    new Roles
                    {
                        Id = index.Id,
                        Name = index.Name,
                        ClientRole = index.ClientRole
                    });

                yield return new InvitedUsers
                {
                    UserId = item.UserId,
                    EmailId = item.EmailId,
                    InvitationStatus = item.InvitationStatus,
                    InvitedUserRoles = invitedUserRoles
                };
            }
        }                

```

src/registration/CatenaX.NetworkServices.Registration.Service/Controllers/RegistrationController.cs
```diff
@@ -301,5 +301,22 @@ public async Task<IActionResult> SubmitRegistrationAsync()
                 return new StatusCodeResult((int)HttpStatusCode.InternalServerError);
             }
         }
+
+        [HttpGet]
+        [Authorize(Roles = "view_registration")]
+        [Route("application/{applicationId}/invitedusers")]
+        public async Task<IActionResult> GetInvitedUserDetailAsync([FromRoute] Guid applicationId)
```
ntruchsess, line 308: I suggest:
```
        [HttpGet]
        [Authorize(Roles = "view_registration")]
        [Route("application/{applicationId}/invitedusers")]
        public IAsyncEnumerable<InvitedUsers> GetInvitedUserDetailAsync([FromRoute] Guid applicationId) =>
            _registrationBusinessLogic.GetInvitedUsersDetail(applicationId);
```
The controller knows how to handle IAsyncEnumerable response type properly. In dotnet6 it will even stream it to the client without buffering the whole response in memory first which allows kinda unlimited size of response.

Regarding the errorhandling-boilerplate code: We already have an errorhandler configured in the middleware, that explicit errorhandling and logging in the controller is now redundant (and will be removed from all constoller-methods in the near future).

ntruchsess (changes requested):

src/portalbackend/CatenaX.NetworkServices.PortalBackend.DBAccess/Models/InvitedUsers.cs
jjeroch: @ntruchsess  & @qxz2mqe  true, we can delete row 10. The UserID is not used in this scenario

src/portalbackend/CatenaX.NetworkServices.PortalBackend.DBAccess/Models/InvitedUsers.cs
jjeroch: We need the Client Role itself. The role id is not relevant, but we need to ensure that we receive the Role Name - to be even more exact => it needs to be the client composite role name

src/portalbackend/CatenaX.NetworkServices.PortalBackend.DBAccess/Models/InvitedUsers.cs
qxz2mqe: fixed

src/portalbackend/CatenaX.NetworkServices.PortalBackend.DBAccess/Models/InvitedUsers.cs
qxz2mqe: fixed

src/registration/CatenaX.NetworkServices.Registration.Service/BusinessLogic/RegistrationBusinessLogic.cs
qxz2mqe: fixed

src/registration/CatenaX.NetworkServices.Registration.Service/Controllers/RegistrationController.cs
qxz2mqe: fixed

src/portalbackend/CatenaX.NetworkServices.PortalBackend.DBAccess/Models/InvitedUsers.cs
ntruchsess: userid removed from response

ntruchsess (approved):

PR #33, opened 2022-05-05
CPLP-618 fix new partner invitation application status CREATED

ntruchsess:
fixes the review-remark by @jjeroch , setting the initial application status to 'CREATED'


evegufy (approved):
Tested successfully.

PR #42, opened 2022-05-10
Feature/cplp 920 user invitation

ntruchsess:
new endpoint POST api/administration/company/{companyId}/users to replace POST api/administration/tenant/{tenant}/users

does update all relvant database-tables (including company_user_assigned_roles)
does validate user calling the endpoint is eligible to do so (role + same company policy)
does validate roles are valid within the db.

unified coding and parameters with existing endpoint POST api/registration/application/{applicationId}/inviteNewUser
does accept additional parameters 'role' and 'message'
does update all relvant database-tables (including company_user_assigned_roles)
does validate user calling the endpoint is eligible to do so (role + same company policy)
does validate roles are valid within the db.

PR depends on database-schema-changes that must be merged in parallel:
https://github.com/tractusx-team-portal-onboarding/devenv/pull/27


src/administration/CatenaX.NetworkServices.Administration.Service/Controllers/UserController.cs
```diff
@@ -33,20 +27,14 @@ public UserController(ILogger<UserController> logger, IUserBusinessLogic logic)
...
-            }
-        }
+        public IAsyncEnumerable<string> ExecuteUserCreation([FromRoute] string tenant, [FromBody] IEnumerable<UserCreationInfo> usersToCreate) =>
+            WithIamUserId(createdByName => _logic.CreateTenantUsersAsync(usersToCreate, tenant, createdByName));
+
+        [HttpPost]
+        [Authorize(Roles = "add_user_account")]
+        [Route("company/{companyId}/users")]
+        public IAsyncEnumerable<string> ExecuteCompanyUserCreation([FromRoute] Guid companyId, [FromBody] IEnumerable<UserCreationInfo> usersToCreate) =>
+            WithIamUserId(createdByName => _logic.CreateCompanyUsersAsync(usersToCreate, companyId, createdByName));
```
jjeroch, line 37: @ntruchsess quick check on the logic of the endpoint. Why do we use the companyId inside the endpoint/route. We could change the logic (which i feel even enhances security) to  company/users....basically ignoring the companyID on the FE. We only take the user token from the acting user, we check which company the user belongs to and add the new user under the same company/idp.

src/provisioning/CatenaX.NetworkServices.Provisioning.Library/Extensions/UserCreationInfo.cs
```diff
@@ -1,17 +1,32 @@
...
+    public UserCreationInfo(string? userName, string email, string? firstName, string? lastName, string? role, string? message)
     {
-        public string userName { get; set; }
+        this.userName = userName;
+        this.eMail = email;
+        this.firstName = firstName;
+        this.lastName = lastName;
+        this.Role = role;
+        this.Message = message;
+    }
```
jjeroch, line 13: @ntruchsess the field is now changing for the FE team from "eMail" to "email" right? If so, we need a change of the Business Logic otherwise that will not work e2e.
Just give me a quick ping if my understanding is right, I can let Martin know that the change will be impacting his business logic

src/administration/CatenaX.NetworkServices.Administration.Service/Controllers/UserController.cs
ntruchsess: The companId is a required input parameter. When implicitly using the companyId of the currently logged in user it would create a tight coupling in between endpoint semantics and authorization. As a consequence one cannot decide to just change the authorization part and allow the endpoint to be called by e.g. a CX Admin.
The Semantics of an endpoint should never depend on the authorizations required.

src/provisioning/CatenaX.NetworkServices.Provisioning.Library/Extensions/UserCreationInfo.cs
ntruchsess: @jjeroch : yes, in changes from eMail to email, but in the one endpoint only. The other endpoint was using 'email' all the time.

ntruchsess:
@wira-mms : I did add you to the list of reviewers as this PR does contain changes on entities and DbContext. We need to decide who shall pull the other persons branch and resove the conflicts. This depends on how mature your changes already are - this PR here is from my point of view ready to merge

src/provisioning/CatenaX.NetworkServices.Provisioning.Library/Extensions/UserCreationInfo.cs
evegufy: @ntruchsess I think like email, also Role should not be nullable, shouldn't it?

ntruchsess: that is not a technical question. Technically there is no requirement to assign a role to a newly invited user, but functionally it certainly it is not entirely useless to do so (@jjeroch )

src/administration/CatenaX.NetworkServices.Administration.Service/Controllers/UserController.cs
ntruchsess: companyId removed from path, it is now read from the db:
/api/administration/user/owncompany/users

src/provisioning/CatenaX.NetworkServices.Provisioning.Library/Extensions/UserCreationInfo.cs
ntruchsess: parameter 'role' changed to 'roles' accepting an array of strings:
```
[
  {
    "userName": "string",
    "email": "string",
    "firstName": "string",
    "lastName": "string",
    "roles": [
      "string"
    ],
    "message": "string"
  }
]
```

ntruchsess:
Assignment of bpn to newly invited users in registration removed.
CompanyId removed from path in user administration
Multiple roles per invite enabled.

wira-mms (approved):

src/administration/CatenaX.NetworkServices.Administration.Service/Controllers/UserController.cs
jjeroch: (y) i like

src/provisioning/CatenaX.NetworkServices.Provisioning.Library/Extensions/UserCreationInfo.cs
jjeroch: thanks, as discussed, for now we proceed with the option "no role"

jjeroch (approved):

evegufy (approved):
Tested successfully. @ntruchsess I'll update the administration secret before merge.

PR #44, opened 2022-05-11
Feature/cplp 458 get pageable list of applications with details

ntruchsess:
implements https://jira.catena-x.net/browse/CPLP-458


evegufy (approved):
@ntruchsess tested successfully

PR #48, opened 2022-05-12
CPLP-968 change tenant-related user endpoints to owncompany/users

ntruchsess:
implements: https://jira.catena-x.net/browse/CPLP-968


evegufy:
I tested all refactored endpoints successfully except the one for the ownUser deletion where I get an 'Object reference not set to an instance of an object.' exception. Let's have a look at that tomorrow together.

ntruchsess:
thanks for testing and discovering a bug (which is fixed now) :+1:

evegufy (approved):
All refactored endpoints were tested successfully.

PR #52, opened 2022-05-16
CPLP-458 return pagination metadata and content in single endpoint

ntruchsess:
As discussed, returns both metadata and content as properties of a single json-object.
Implementation based on a reusable generic implementation placed in new Module CatenaX.NetworkServices.Framework.Models


evegufy (approved):
Looks good!
@ntruchsess please have a look at before merging, I resolved some conficts.

PR #56, opened 2022-05-18
CPLP-999  password reset adapt  new logic

qxz2mqe:
https://jira.catena-x.net/browse/CPLP-999


src/administration/CatenaX.NetworkServices.Administration.Service/BusinessLogic/UserBusinessLogic.cs
```diff
@@ -330,23 +331,27 @@ private async Task<bool> CanResetPassword(string userId)
...
+                        }
+                        return updatedPassword;
                     }
-                    return updatedPassword;
+                    throw new NotFoundException($"no shared idp user {companyUserId} found in company of {adminUserId}");
                 }
                 throw new ArgumentException($"cannot reset password more often than {_settings.PasswordReset.MaxNoOfReset} in {_settings.PasswordReset.NoOfHours} hours");
             }
-            throw new NotFoundException($"no shared idp user {companyUserId} found in company of {adminUserId}");
+            throw new ArgumentException($"Company ID  does not exist");
```
ntruchsess, line 354: that message should give more information about the root cause - like '$"Cannot identify companyId: companyUserId {companyUserId} is not associated with the same company as adminUserId {adminUserId}"

src/portalbackend/CatenaX.NetworkServices.PortalBackend.DBAccess/PortalBackendDBAccess.cs
```diff
@@ -468,11 +468,16 @@ public async IAsyncEnumerable<WelcomeEmailData> GetWelcomeEmailDataUntrackedAsyn
             _dbContext.IamUsers
                 .Where(iamUser => iamUser.UserEntityId == adminUserId)
```
ntruchsess, line 469: That query certainly works. As an optimization I suggest to include the 'same company'-condition into the very first Where() so the overall result would be null if the condition is not met. That saves the additional steps of data-retrieval from the database (like idp-data) that would be redundant in this case. And don't forget the 'AsNoTracking()' as you are not going to modify the entities being returned.
e.g.:
            _dbContext.CompanyUsers
                .AsNoTracking()
                .Where(companyUser => 
                    companyUser.Id == companyUserId
                    && companyUser.Company!.CompanyUsers.Any(companyUser => companyUser.IamUser!.UserEntityId == adminUserId))
                .Select(companyUser => new IdpUser
                {
                    TargetIamUserId = companyUser.IamUser!.UserEntityId,
                    CompanyId = companyUser.CompanyId,
                    IdpName = companyUser.Company!.IdentityProviders
                        .Where(identityProvider => identityProvider.IdentityProviderCategoryId == IdentityProviderCategoryId.KEYCLOAK_SHARED)
                        .Select(identityProvider => identityProvider.IamIdentityProvider!.IamIdpAlias)
                        .SingleOrDefault()
                }).SingleOrDefaultAsync();

ntruchsess (changes requested):

src/administration/CatenaX.NetworkServices.Administration.Service/BusinessLogic/UserBusinessLogic.cs
qxz2mqe: fixed

src/portalbackend/CatenaX.NetworkServices.PortalBackend.DBAccess/PortalBackendDBAccess.cs
qxz2mqe: fixed

src/administration/CatenaX.NetworkServices.Administration.Service/BusinessLogic/UserBusinessLogic.cs
```diff
@@ -330,23 +331,27 @@ private async Task<bool> CanResetPassword(string userId)
...
         {
             var idpUserName = await _portalDBAccess.GetIdpCategoryIdByUserId(companyUserId, adminUserId).ConfigureAwait(false);
-            if (idpUserName?.IdpName == tenant && !string.IsNullOrWhiteSpace(idpUserName?.TargetIamUserId))
+            if (idpUserName?.CompanyId != null)
             {
                 if (await CanResetPassword(adminUserId).ConfigureAwait(false))
                 {
-                    var updatedPassword = await _provisioningManager.ResetSharedUserPasswordAsync(tenant, idpUserName.TargetIamUserId).ConfigureAwait(false);
-                    if (!updatedPassword)
+                    if (!string.IsNullOrWhiteSpace(idpUserName?.TargetIamUserId) && !string.IsNullOrWhiteSpace(idpUserName?.IdpName))
```
ntruchsess, line 341: this if-statement should be executed before the call to CanResetPassword as the reset cannot be done if  targetIamUserid and/or idpName are not known so there is no point in updating the database in this case.

ntruchsess (changes requested):

src/administration/CatenaX.NetworkServices.Administration.Service/BusinessLogic/UserBusinessLogic.cs
qxz2mqe: fixed

ntruchsess (approved):

PR #58, opened 2022-05-19
Feature/cplp 629 api get call to retrieve uploaded documents

Akshay-Chaudhari-BMW:
api get call to retrieve uploaded documents

### DIFF
--- a/src/registration/CatenaX.NetworkServices.Registration.Service/BusinessLogic/RegistrationBusinessLogic.cs
+++ b/src/registration/CatenaX.NetworkServices.Registration.Service/BusinessLogic/RegistrationBusinessLogic.cs
@@ -388,15 +388,16 @@ namespace CatenaX.NetworkServices.Registration.Service.BusinessLogic
                 );
             }
         }
+        //TODO: Need to implement storage for document upload
         public async IAsyncEnumerable<UploadDocumentDetails> GetUploadedDocumentsAsync(Guid applicationId, DocumentTypeId documentTypeId, string iamUserId)
         {
             await foreach (var documents in _portalDBAccess.GetUploadedDocumentsAsync(applicationId,documentTypeId,iamUserId).ConfigureAwait(false))
             {
                 yield return new UploadDocumentDetails
-                {
-                    DocumentId = documents.DocumentId,
-                    DocumentName = documents.DocumentName
-                };
+                (
+                    documents.DocumentId,
+                    documents.DocumentName
+                );
             }
         }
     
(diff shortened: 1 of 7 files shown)


src/registration/CatenaX.NetworkServices.Registration.Service/BusinessLogic/RegistrationBusinessLogic.cs
```diff
@@ -388,5 +388,18 @@ await foreach (var item in _portalDBAccess.GetInvitedUserDetailsUntrackedAsync(a
                 );
             }
         }
+        //TODO: Need to implement storage for document upload
+        public async IAsyncEnumerable<UploadDocumentDetails> GetUploadedDocumentsAsync(Guid applicationId, DocumentTypeId documentTypeId, string iamUserId)
```
ntruchsess, line 392: actually it's just fine to return the db-model as-is, there is no benefit in re-wrapping it into an object that has exactly the very same properties.

ntruchsess (changes requested):

src/registration/CatenaX.NetworkServices.Registration.Service/BusinessLogic/RegistrationBusinessLogic.cs
Akshay-Chaudhari-BMW: model replace with db-model

src/registration/CatenaX.NetworkServices.Registration.Service/BusinessLogic/RegistrationBusinessLogic.cs
```diff
@@ -388,5 +388,18 @@ await foreach (var item in _portalDBAccess.GetInvitedUserDetailsUntrackedAsync(a
                 );
             }
         }
+        //TODO: Need to implement storage for document upload
+        public async IAsyncEnumerable<UploadDocuments> GetUploadedDocumentsAsync(Guid applicationId, DocumentTypeId documentTypeId, string iamUserId)
```
ntruchsess, line 392: as being said: no need to re-wrap. Just do:
```
        public IAsyncEnumerable<UploadDocuments> GetUploadedDocumentsAsync(Guid applicationId, DocumentTypeId documentTypeId, string iamUserId) =>
            _portalDBAccess.GetUploadedDocumentsAsync(applicationId,documentTypeId,iamUserId);

```

ntruchsess (changes requested):

src/registration/CatenaX.NetworkServices.Registration.Service/BusinessLogic/RegistrationBusinessLogic.cs
Akshay-Chaudhari-BMW: resolved

ntruchsess (approved):

PR #60, opened 2022-05-20
CPLP-786 Connector API groundwork: database extensions

wira-mms:


src/portalbackend/CatenaX.NetworkServices.PortalBackend.PortalEntities/Entities/Connector.cs
```diff
@@ -0,0 +1,40 @@
...
+public class Connector
+{
+    public Connector(string name, string locationId, string connectorUrl)
+    {
+        Name = name;
+        LocationId = locationId;
+        ConnectorUrl = connectorUrl;
+    }
+
+    public Guid Id { get; set; }
```
ntruchsess, line 15: the setter of the key-field should be private

ntruchsess:
For better overview there should be single new Migration per PR, nobody will want to find out which intermediate commit this second Migration is referring to.

ntruchsess (changes requested):

src/portalbackend/CatenaX.NetworkServices.PortalBackend.PortalEntities/Entities/Connector.cs
wira-mms: https://github.com/catenax-ng/product-portal-backend/pull/60/commits/e2abce43c323d399b9d7e6b9fea4afb29c4924af

wira-mms:
> For better overview there should be a single new Migration per PR, nobody will want to find out which intermediate commit this second Migration is referring to. Just do 'dotnet ef migrations remove' two times followed by a single 'dotnet ef migrations add xxx'

Tagged migrations in git history

ntruchsess (approved):

PR #63, opened 2022-05-24
CPLP-912 implement new endpoint 'ownUser' to retrieve and update own user data

ntruchsess:
implemented new endpoints to read and update own user details.
@oyo : details in Userstory: https://jira.catena-x.net/browse/CPLP-912


evegufy (approved):
@ntruchsess Looks good! Tested successfully.

PR #71, opened 2022-05-31
Feature/cplp 1022 update invite status

Akshay-Chaudhari-BMW:
update invite status after login


src/registration/CatenaX.NetworkServices.Registration.Service/BusinessLogic/RegistrationBusinessLogic.cs
```diff
@@ -388,5 +389,21 @@ await foreach (var item in _portalDBAccess.GetInvitedUserDetailsUntrackedAsync(a
         //TODO: Need to implement storage for document upload
         public IAsyncEnumerable<UploadDocuments> GetUploadedDocumentsAsync(Guid applicationId, DocumentTypeId documentTypeId, string iamUserId) =>
             _portalDBAccess.GetUploadedDocumentsAsync(applicationId,documentTypeId,iamUserId);
+
+        public async Task<int> SetInvitationStatusAsync(string iamUserId)
+        {
+            var invitationData = await _portalDBAccess.GetInvitationStatusAsync(iamUserId).ConfigureAwait(false);
+            
+                if (invitationData.InvitationStatusId == InvitationStatusId.CREATED)
```
ntruchsess, line 397: Please stop ignoring the nullable warnings. There is a good reason for that, code should never ever again throw NullReferenceException

ntruchsess (changes requested):

src/registration/CatenaX.NetworkServices.Registration.Service/BusinessLogic/RegistrationBusinessLogic.cs
Akshay-Chaudhari-BMW: handle this case with exception.

src/registration/CatenaX.NetworkServices.Registration.Service/BusinessLogic/RegistrationBusinessLogic.cs
```diff
@@ -389,6 +390,27 @@ await foreach (var item in _portalDBAccess.GetInvitedUserDetailsUntrackedAsync(a
...
+                throw new ForbiddenException($"iamUserId {iamUserId} is not associated with invitation");
+            }
+            
+            if (invitationData.InvitationStatusId == InvitationStatusId.CREATED)
+            {
+                invitationData.InvitationStatusId = InvitationStatusId.PENDING;
+            }
+            else
+            {
+                return (int)HttpStatusCode.AlreadyReported;
```
ntruchsess, line 408: returning int as payload will not result in Statuscode 208. You would have to add a new Exception-class to src/framework/CatenaX.NetworkServices.Framework.ErrorHandling and add the code catching this exception transforming it into HttpStatusCode.AlreadyReported in Class GeneralHttpErrorHandler.
But anyway - returning 208 is inappropriate as this is not a WebDAV-request. See my comment in https://jira.catena-x.net/browse/CPLP-1022 and negotiate with @jjeroch whether this should result in 200 OK or 400 Bad Request. (To return 400 Bad Request you'd throw ArgumentException)

ntruchsess (changes requested):

src/registration/CatenaX.NetworkServices.Registration.Service/BusinessLogic/RegistrationBusinessLogic.cs
jjeroch: As discussed - 400 :) but with an error message

Akshay-Chaudhari-BMW: handle with 400 - bad request

PR #78, opened 2022-06-03
CPLP-1050 App subscription database additions

wira-mms:


ntruchsess (changes requested):
as a conflicting migration has been created in parallel please merge main and re-create the migration of this PR

ntruchsess (approved):
:+1:

PR #82, opened 2022-06-08
Cplp 464 get api to show app roles

Akshay-Chaudhari-BMW:
get api to show app roles

### DIFF
--- a/src/administration/CatenaX.NetworkServices.Administration.Service/Controllers/UserController.cs
+++ b/src/administration/CatenaX.NetworkServices.Administration.Service/Controllers/UserController.cs
@@ -67,9 +67,9 @@ namespace CatenaX.NetworkServices.Administration.Service.Controllers
 
         [HttpGet]
         [Authorize(Roles = "view_client_roles")]
-        [Route("app/{applicationId}/roles")]
-        public IAsyncEnumerable<ClientRoles> GetClientRolesAsync([FromRoute] Guid applicationId,string? languageShortName = null) =>
-            _logic.GetClientRolesAsync(applicationId,languageShortName);
+        [Route("app/{appId}/roles")]
+        public IAsyncEnumerable<ClientRoles> GetClientRolesAsync([FromRoute] Guid appId,string? languageShortName = null) =>
+            _logic.GetClientRolesAsync(appId,languageShortName);
 
         [HttpGet]
         [Route("ownUser")]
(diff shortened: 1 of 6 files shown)


src/administration/CatenaX.NetworkServices.Administration.Service/Controllers/UserController.cs
```diff
@@ -67,9 +67,9 @@ public UserController(ILogger<UserController> logger, IUserBusinessLogic logic)
 
         [HttpGet]
         [Authorize(Roles = "view_client_roles")]
-        [Route("client/{clientId}/roles")]
-        public Task<IEnumerable<string>> ReturnRoles([FromRoute] string clientId) =>
-            _logic.GetAppRolesAsync(clientId);
+        [Route("app/{applicationId}/roles")]
```
jjeroch, line 70: @Akshay-Chaudhari-BMW  can you please change this to _appID_ instead of _applicationID_
"Application" is used for the company applications and app for the services. It would be confusing if we call it "applicationId"

src/administration/CatenaX.NetworkServices.Administration.Service/Controllers/UserController.cs
```diff
@@ -67,9 +67,9 @@ public UserController(ILogger<UserController> logger, IUserBusinessLogic logic)
...
+        public IAsyncEnumerable<ClientRoles> GetClientRolesAsync([FromRoute] Guid applicationId,string? languageShortName = null) =>
+            _logic.GetClientRolesAsync(applicationId,languageShortName);
```
jjeroch, line 72: Same here, please dont call it applicationID

jjeroch (dismissed):

src/administration/CatenaX.NetworkServices.Administration.Service/Controllers/UserController.cs
Akshay-Chaudhari-BMW: replaced applicationId with appId

src/administration/CatenaX.NetworkServices.Administration.Service/Controllers/UserController.cs
Akshay-Chaudhari-BMW: replaced applicationId with appId

evegufy (changes requested):
@Akshay-Chaudhari-BMW as just discussed: Please use the word 'app' everywhere where you really mean app and not application like in the sense of becoming member of the consortia.

evegufy:
@Akshay-Chaudhari-BMW @jjeroch If there are additional endpoints coming up for the administration of apps we may want to establish an own controller (and logic and so on) for Apps, not having it under the User controller anymore as it doesn't really fit there functionally (technically it's not an issue to leave it in the User Controller). I wouldn't do that for one endpoint, just to have that in mind for an upcoming user stories or sub tasks.

Akshay-Chaudhari-BMW:
> @Akshay-Chaudhari-BMW as just discussed: Please use the word 'app' everywhere where you really mean app and not application like in the sense of becoming member of the consortia.

@evegufy done

evegufy (approved):

PR #89, opened 2022-06-10
cplp -1022 - remove forbidden for no status update.

Akshay-Chaudhari-BMW:
remove forbidden for no status update.


jjeroch:
@Akshay-Chaudhari-BMW  I can see that you removed the exception. However doesn't this result now still in an issue if the invitation status is PENDING, ACCEPTED and DECLINED

jjeroch (commented):
@Akshay-Chaudhari-BMW I can see that you removed the exception. However doesn't this result now still in an issue if the invitation status is PENDING, ACCEPTED and DECLINED

Akshay-Chaudhari-BMW:
> @Akshay-Chaudhari-BMW I can see that you removed the exception. However doesn't this result now still in an issue if the invitation status is PENDING, ACCEPTED and DECLINED

@jjeroch If there is no status update then response will be empty body.

jjeroch:
As just discussed we update it as following "if status is "2" (PENDING) we will update the invitation status to "3" (ACCEPTED); if status is "1", "3" or "4" we send back an empty body"

Akshay-Chaudhari-BMW:
Invitation status update done

jjeroch (approved):
Looks good, approved

PR #98, opened 2022-06-22
CPLP-1040 reimplement double user check fixing CPLP-1116

ntruchsess:
implements fix for https://jira.catena-x.net/browse/CPLP-1116

initial implementation of CPLP-1040 invite user double user check ( https://jira.catena-x.net/browse/CPLP-1040 ) was actually checking whether the logged in user (from the token) already exists in table iam_users. As long the currently logged in user was created accordingly this condition will always be true.
New implementation as of this PR checks whether in the company of the currently logged in user a user with the same email-address as the to-be-created user already does exist.
(new implementation includes refactoring of business-method to repository-pattern)


Phil91 (approved):
Looks fine by me.

PR #102, opened 2022-06-23
CPLP-1104 batch delete

Phil91:
✅ Updated KeePass
✅ Added Maintenance Service
✅ Delete Documents and depending consents older x days (x is configurable in the appSettings)

### DIFF
--- a/src/maintenance/CatenaX.NetworkServices.Maintenance.App/BatchDeleteService.cs
+++ b/src/maintenance/CatenaX.NetworkServices.Maintenance.App/BatchDeleteService.cs
@@ -1,30 +1,52 @@
 ﻿using CatenaX.NetworkServices.PortalBackend.PortalEntities;
 using Microsoft.EntityFrameworkCore;
 
-namespace CatenaX.NetworkServices.Maintenance.App
+namespace CatenaX.NetworkServices.Maintenance.App;
+
+/// <summary>
+/// Service to delete the pending and inactive documents as well as the depending consents from the database
+/// </summary>
+public class BatchDeleteService : BackgroundService
 {
+    private readonly IServiceScopeFactory _serviceScopeFactory;
+    private readonly ILogger<BatchDeleteService> _logger;
+    private readonly int _days;
+
     /// <summary>
-    /// Service to delete the pending and inactive documents from the database
+    /// Creates a new instance of <see cref="BatchDeleteService"/>
     /// </summary>
-    public static class BatchDeleteService
+    /// <param name="serviceScopeFactory">access to the services</param>
+    /// <param name="logger">the logger</param>
+    /// <param name="config">the apps configuration</param>
+    public BatchDeleteService(IServiceScopeFactory serviceScopeFactory, ILogger<BatchDeleteService> logger, IConfiguration config)
     {
-        /// <summary>
-        /// Deletes the inactive and pending documents that are older than the specified days
-        /// </summary>
-        /// <param name="dbContext"></param>
-        /// <param name="days"></param>
-        /// <exception cref="Exception"></exception>
-        public static void BatchDeleteDocuments(PortalDbContext dbContext, int days)
+        _serviceScopeFactory = serviceScopeFactory;
+        _logger = logger;
+        _days = config.GetValue<int>("DeleteIntervalInDays");
+    }
+
+    /// <inheritdoc />
+    protected override Task ExecuteAsync(CancellationToken stoppingToken)
+    {
+        using var scope = _serviceScopeFactory.CreateScope();
+        var dbContext = scope.ServiceProvider.GetRequiredService<PortalDbContext>();
+            
+        using var transaction = dbContext.Database.BeginTransaction();
+        try
         {
-            try
-            {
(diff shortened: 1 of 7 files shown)


src/maintenance/CatenaX.NetworkServices.Maintenance.App/BatchDeleteService.cs
```diff
@@ -0,0 +1,49 @@
...
+    protected override Task ExecuteAsync(CancellationToken stoppingToken)
+    {
+        using var scope = _serviceScopeFactory.CreateScope();
+        var dbContext = scope.ServiceProvider.GetRequiredService<PortalDbContext>();
+            
+        try
+        {
+            _logger.LogInformation($"Cleaning up documents and consents older {_days} days...");
+            var dbQuery = string.Format("WITH documentids AS (DELETE FROM portal.documents WHERE date_created < now() - interval '{0} days' and (document_status_id = {1} or document_status_id = {2}) RETURNING id) DELETE FROM portal.consents WHERE document_id in (SELECT id from documentids);", _days, (int) DocumentStatusId.PENDING, (int) DocumentStatusId.INACTIVE);
+            dbContext.Database.ExecuteSqlRaw(dbQuery);
```
ntruchsess, line 39: Please don't use string.format - there is an overload of ExecuteSqlRaw that will take the query-string that contains the positional parameters and translate that into a sql with parameters, passing the parameter-values not within the query but separately. As an alternative use ExecuteSqlInterpolated instead
Use the async-version of those methods, awaiting them and passing the cancellationToken (parameter stoppingToken of ExecuteAsync) (seems we need to find out how to enable additional Workers to run this).

ntruchsess (changes requested):

src/maintenance/CatenaX.NetworkServices.Maintenance.App/BatchDeleteService.cs
Phil91: fixed it to ExecuteSqlInterpolatedAsync. The worker is running asynchronously now as well

src/maintenance/CatenaX.NetworkServices.Maintenance.App/BatchDeleteService.cs
```diff
@@ -0,0 +1,59 @@
...
+        _days = config.GetValue<int>("DeleteIntervalInDays");
+    }
+
+    /// <inheritdoc />
+    protected override async Task ExecuteAsync(CancellationToken stoppingToken)
+    {
+        using var scope = _serviceScopeFactory.CreateScope();
+        var dbContext = scope.ServiceProvider.GetRequiredService<PortalDbContext>();
+            
+        while (!_applicationLifetime.ApplicationStopping.IsCancellationRequested)
```
ntruchsess, line 42: from my understanding this should be 'if(!stoppingToken.IsCancellationRequestet)' instead of querying the applicationLifetime.
The 'while' would cause this task to run continiously until cancelled - from my understanding we did intend to do the scheduling by kubernetes cron-job, so this should run once per job execution (if we were not calling _applicationLifetime.StopApplication() at the end).

ntruchsess (changes requested):

src/maintenance/CatenaX.NetworkServices.Maintenance.App/BatchDeleteService.cs
```diff
@@ -0,0 +1,59 @@
...
+            await dbContext.Database.ExecuteSqlInterpolatedAsync($"WITH documentids AS (DELETE FROM portal.documents WHERE date_created < {DateTimeOffset.UtcNow.AddDays(-_days)} AND (document_status_id = {(int)DocumentStatusId.PENDING} OR document_status_id = {(int) DocumentStatusId.INACTIVE}) RETURNING id) DELETE FROM portal.consents WHERE document_id IN (SELECT id FROM documentids);", stoppingToken).ConfigureAwait(false);
+            _logger.LogInformation($"Documents older than {_days} days and depending consents successfully cleaned up.");
+        }
+        catch (Exception ex)
+        {
+            _logger.LogError($"Database clean up failed with error: {ex.Message}");
+        }
+
+        Environment.ExitCode = 1;
+        if (!stoppingToken.IsCancellationRequested)
```
ntruchsess, line 54: I suggest putting this 'if' above line 42 (the 'try'), and the the assignment of ExitCode below line 57 (the end of the if-block)

ntruchsess (changes requested):

src/maintenance/CatenaX.NetworkServices.Maintenance.App/BatchDeleteService.cs
Phil91: but with moving the StopApplication above the actually cleanup, the cleanup would never be processed right?

Phil91: i tried to move the if statement above the try, but it didn't work. the database context can't connect to the database when trying to delete the documents.

ntruchsess (approved):

PR #104, opened 2022-06-27
CPLP-466 fix handling of inconsistent keycloak data

ntruchsess:
fixing https://jira.catena-x.net/browse/CPLP-466
RoleManager.AssignClientRolesToCentralUserAsync would not throw NotFoundException if roles or clients are not existing in keycloak. All places where this is called by the business-logic were adjusted accordingly. PR also includes some refactoring to repository-pattern as I made sure the methods I had to touch anyway no longer use the PortalDBAccess-class any more.


ntruchsess:
@jjeroch I see your proposal in https://jira.catena-x.net/browse/CPLP-1115 results in code that will cause trouble in the long-run as it tries to deal with both scenarios (single role vs. multiple roles to be assigned) within a single algorithm. Please see my comment in CPLP-1115

src/provisioning/CatenaX.NetworkServices.Provisioning.Library/Extensions/RoleManager.cs
```diff
@@ -8,36 +10,30 @@ private async Task<(string, IEnumerable<Role>)> GetCentralClientIdRolesAsync(str
...
-            if (client == null)
-            {
-                throw new Exception($"failed to retrieve central client {clientName}");
-            }
             switch (count)
             {
                 case 0:
                     return (client.Id, Enumerable.Empty<Role>());
                 case 1:
                     return (client.Id, Enumerable.Repeat<Role>(await _CentralIdp.GetRoleByNameAsync(_Settings.CentralRealm, client.Id, roleNames.Single()).ConfigureAwait(false), 1));
```
ntruchsess, line 18: GetRoleByNameAsync may throw NotFoundException. To guarantee consistent behaviour this must be caught and and (client.id, Enumerable.Empty) must be returned.

src/provisioning/CatenaX.NetworkServices.Provisioning.Library/Extensions/RoleManager.cs
```diff
@@ -8,36 +10,30 @@ private async Task<(string, IEnumerable<Role>)> GetCentralClientIdRolesAsync(str
...
+                    await _CentralIdp.AddClientRoleMappingsToUserAsync(_Settings.CentralRealm, centralUserId, clientId, roles).ConfigureAwait(false);
+                    return roles;
+                    
                 }
-            )).ConfigureAwait(false)).All(x => x))
-            {
-                throw new Exception($"failed to assign client-roles {clientRoleNames} to central user {centralUserId}");
-            }
+            )).ConfigureAwait(false);
+            return results.Select(role => role.FirstOrDefault().Name);
```
ntruchsess, line 36: that does not make a lot of sense to just return the name of the first role that has been assigned successful as the information of the other roles that have been assigned is lost.
If you implement my proposal in line 32 comment result will be (string, IEnumerable<string>)[] which could easily be transformed into a dictionary so we would return the same type as we get as input-parameter (just containing less roles in case some were not assigned).
This way the businesslogic can compare the roles that were passed as input-parameter with the roles that have actually been assigned in a consistent way.
The decision whether to throw any exception or return an ok-result then would be implemented entirely within the business-logic which would be way better maintainable as that logic would be in a single place and the methods that are being called would behave consistently (not throwing NotFoundException in any case)

src/provisioning/CatenaX.NetworkServices.Provisioning.Library/Extensions/RoleManager.cs
```diff
@@ -8,36 +10,30 @@ private async Task<(string, IEnumerable<Role>)> GetCentralClientIdRolesAsync(str
...
-        public async Task AssignClientRolesToCentralUserAsync(string centralUserId, IDictionary<string, IEnumerable<string>> clientRoleNames)
+        public async Task<IEnumerable<string>> AssignClientRolesToCentralUserAsync(string centralUserId, IDictionary<string, IEnumerable<string>> clientRoleNames)
         {
-            if (!(await Task.WhenAll(clientRoleNames.Select(async x =>
+            var results = await Task.WhenAll(clientRoleNames.Select(async x =>
                 {
                     var (client, roleNames) = x;
                     var (clientId, roles) = await GetCentralClientIdRolesAsync(client, roleNames).ConfigureAwait(false);
-                    return await _CentralIdp.AddClientRoleMappingsToUserAsync(_Settings.CentralRealm, centralUserId, clientId, roles).ConfigureAwait(false);
+                    await _CentralIdp.AddClientRoleMappingsToUserAsync(_Settings.CentralRealm, centralUserId, clientId, roles).ConfigureAwait(false);
```
ntruchsess, line 31: if you implement my proposal to return null for the clientId in case the client is not found that would have to be checked here before calling Add ClientRoleMappingsToUserAsync

src/provisioning/CatenaX.NetworkServices.Provisioning.Library/Extensions/RoleManager.cs
```diff
@@ -8,36 +10,30 @@ private async Task<(string, IEnumerable<Role>)> GetCentralClientIdRolesAsync(str
         {
             var count = roleNames.Count();
             var client = await GetCentralClientViewableAsync(clientName).ConfigureAwait(false);
```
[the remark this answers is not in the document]
qxz2mqe: fixed

src/provisioning/CatenaX.NetworkServices.Provisioning.Library/Extensions/RoleManager.cs
qxz2mqe: fixed

src/provisioning/CatenaX.NetworkServices.Provisioning.Library/Extensions/RoleManager.cs
```diff
@@ -8,36 +10,30 @@ private async Task<(string, IEnumerable<Role>)> GetCentralClientIdRolesAsync(str
...
+        public async Task<IEnumerable<string>> AssignClientRolesToCentralUserAsync(string centralUserId, IDictionary<string, IEnumerable<string>> clientRoleNames)
         {
-            if (!(await Task.WhenAll(clientRoleNames.Select(async x =>
+            var results = await Task.WhenAll(clientRoleNames.Select(async x =>
                 {
                     var (client, roleNames) = x;
                     var (clientId, roles) = await GetCentralClientIdRolesAsync(client, roleNames).ConfigureAwait(false);
-                    return await _CentralIdp.AddClientRoleMappingsToUserAsync(_Settings.CentralRealm, centralUserId, clientId, roles).ConfigureAwait(false);
+                    await _CentralIdp.AddClientRoleMappingsToUserAsync(_Settings.CentralRealm, centralUserId, clientId, roles).ConfigureAwait(false);
+                    return roles;
```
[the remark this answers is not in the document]
qxz2mqe: fixed

src/provisioning/CatenaX.NetworkServices.Provisioning.Library/Extensions/RoleManager.cs
qxz2mqe: fixed

src/provisioning/CatenaX.NetworkServices.Provisioning.Library/Extensions/RoleManager.cs
qxz2mqe: fixed

src/provisioning/CatenaX.NetworkServices.Provisioning.Library/Extensions/RoleManager.cs
```diff
@@ -7,37 +10,54 @@ public partial class ProvisioningManager
         private async Task<(string, IEnumerable<Role>)> GetCentralClientIdRolesAsync(string clientName, IEnumerable<string> roleNames)
         {
             var count = roleNames.Count();
-            var client = await GetCentralClientViewableAsync(clientName).ConfigureAwait(false);
-            if (client == null)
+            Client client = null;
```
ntruchsess, line 13: that must be declared as nullable:
`Client? client = null;`

src/provisioning/CatenaX.NetworkServices.Provisioning.Library/Extensions/RoleManager.cs
ntruchsess: - I cannot see where `clientId == null` is being checked? AddClientRoleMappingsToUserAsync will throw Object is null reference exception when passing null als clientId
- the return-value of AddClientRoleMappingsToUserAsync must be taken into account as whenever role-assignements fail for other reasons those roles must not be included in the list of successful assigned roles

src/provisioning/CatenaX.NetworkServices.Provisioning.Library/Extensions/RoleManager.cs
```diff
@@ -7,37 +10,54 @@ public partial class ProvisioningManager
...
+                        return (null, Enumerable.Empty<Role>());
                     }
+                    return (client.Id, Enumerable.Empty<Role>());
+                }
             }
+            return (null, Enumerable.Empty<Role>());
         }
 
-        public async Task AssignClientRolesToCentralUserAsync(string centralUserId, IDictionary<string, IEnumerable<string>> clientRoleNames)
+        public async Task<IDictionary<string, IEnumerable<string>>[]> AssignClientRolesToCentralUserAsync(string centralUserId, IDictionary<string, IEnumerable<string>> clientRoleNames)
```
ntruchsess, line 44: what ist the Array good for? Return-type should be Task<IDictionary<string, IEnumerable<string>>>

src/provisioning/CatenaX.NetworkServices.Provisioning.Library/Extensions/RoleManager.cs
```diff
@@ -7,37 +10,54 @@ public partial class ProvisioningManager
...
                     var (clientId, roles) = await GetCentralClientIdRolesAsync(client, roleNames).ConfigureAwait(false);
-                    return await _CentralIdp.AddClientRoleMappingsToUserAsync(_Settings.CentralRealm, centralUserId, clientId, roles).ConfigureAwait(false);
+                    await _CentralIdp.AddClientRoleMappingsToUserAsync(_Settings.CentralRealm, centralUserId, clientId, roles).ConfigureAwait(false);
+                    IEnumerable<string> rolesList = roleNames.Except(roles.Select(role => role.Name));
+                    return  (client:client, rolesList:rolesList);
+                    
                 }
-            )).ConfigureAwait(false)).All(x => x))
+            )).ConfigureAwait(false);
+            var assignedClientRoles = new Dictionary<string, IEnumerable<string>>[]
```
ntruchsess, line 56: what ist the Array good for? It will always contain a single element, there cannot be more.

src/provisioning/CatenaX.NetworkServices.Provisioning.Library/Extensions/RoleManager.cs
```diff
@@ -7,37 +10,54 @@ public partial class ProvisioningManager
...
+                    return  (client:client, rolesList:rolesList);
+                    
                 }
-            )).ConfigureAwait(false)).All(x => x))
+            )).ConfigureAwait(false);
+            var assignedClientRoles = new Dictionary<string, IEnumerable<string>>[] 
             {
-                throw new Exception($"failed to assign client-roles {clientRoleNames} to central user {centralUserId}");
-            }
+              clientRoles.ToDictionary(clientRole => clientRole.client, clientRole => clientRole.rolesList)
```
ntruchsess, line 58: you need to exclude the items where clientId is null:
`return clientRoles.Where(clientRole => clientRole.client != null).ToDictionary(clientRole => clientRole.client, clientRole => clientRole.rolesList);`

src/administration/CatenaX.NetworkServices.Administration.Service/BusinessLogic/UserBusinessLogic.cs
```diff
@@ -511,24 +511,19 @@ public async Task<string> AddUserRoleAsync(Guid appId, UserRoleInfo userRoleInfo
...
-
-                try
+                var roleList = await _provisioningManager.AssignClientRolesToCentralUserAsync(companyUser.TargetIamUserId, clientRoleNames).ConfigureAwait(false);
+                isApiCallSuccess = true;
+                foreach (Dictionary<string, IEnumerable<string>> kvp in roleList)
                 {
-                    var roleList = await _provisioningManager.AssignClientRolesToCentralUserAsync(companyUser.TargetIamUserId, clientRoleNames).ConfigureAwait(false);
-                    isApiCallSuccess = true;
-                    foreach (var item in roles)
+                    KeyValuePair<string, IEnumerable<string>> outputRole = kvp.ElementAt(0);
```
ntruchsess, line 518: - use 'Single()' instead of 'ElementAt(0)'. That is explicitly saying that exactly one client is expected in the result of AssignClientRolesToCentralUserAsync and having more than a single client in the dictionary here is not expected. That is important as the method eventually will be changed to accept multiple clients and if this is not being adjusted accordingly it shall not ignore this silently.
- a KeyValuePair can be directly deconstructed to get reasonable named variables instead of the generic ones.
```
var (client, outputRoles) = roleList[0].Single();
```

src/administration/CatenaX.NetworkServices.Administration.Service/BusinessLogic/UserBusinessLogic.cs
```diff
@@ -511,24 +511,19 @@ public async Task<string> AddUserRoleAsync(Guid appId, UserRoleInfo userRoleInfo
...
+                    if (outputRole.Value.Any())
                     {
-                        if (!roleList.Contains(item))
-                        {
-                            message.Append($"Warning- {item} failed to get assigned. Please contact the service team.");
-                        }
+                        isApiCallSuccess = false;
+                        var outputRoles = outputRole.Value.Select(x => x);
```
ntruchsess, line 522: this line is a no-op and redundant. outputRoles will have the same elements as outputRole.Value. If you use KeyValuePair-deconstruction as pointed out in my previous comment you can get the named var by direct assignment.

src/administration/CatenaX.NetworkServices.Administration.Service/BusinessLogic/UserBusinessLogic.cs
```diff
@@ -511,24 +511,19 @@ public async Task<string> AddUserRoleAsync(Guid appId, UserRoleInfo userRoleInfo
...
+                        var roleName = string.Join(" ", outputRoles.ToArray());
```
ntruchsess, line 523: ToArray() is redundant as String.Join accepts IEnumerable as parameter-type

src/administration/CatenaX.NetworkServices.Administration.Service/BusinessLogic/UserBusinessLogic.cs
```diff
@@ -511,24 +511,19 @@ public async Task<string> AddUserRoleAsync(Guid appId, UserRoleInfo userRoleInfo
             bool isApiCallSuccess = false;
             if (roles.Count() > 0)
             {
-
-                try
+                var roleList = await _provisioningManager.AssignClientRolesToCentralUserAsync(companyUser.TargetIamUserId, clientRoleNames).ConfigureAwait(false);
```
ntruchsess, line 514: be aware thare are 6 more references to AssignClientRolesToCentralUserAsync that need to be checked and adjusted accordingly. (catch of NotFoundException must be changed to explicit check of result)

src/provisioning/CatenaX.NetworkServices.Provisioning.Library/Extensions/RoleManager.cs
```diff
@@ -7,37 +10,54 @@ public partial class ProvisioningManager
...
                     return (client.Id, Enumerable.Empty<Role>());
                 case 1:
                     return (client.Id, Enumerable.Repeat<Role>(await _CentralIdp.GetRoleByNameAsync(_Settings.CentralRealm, client.Id, roleNames.Single()).ConfigureAwait(false), 1));
                 default:
+                     roles = (await _CentralIdp.GetRolesAsync(_Settings.CentralRealm, client.Id).ConfigureAwait(false)).Where(x => roleNames.Contains(x.Name));
+                     return (client.Id, roles);
+            }
+            
+            }
+            catch(Exception ex)
```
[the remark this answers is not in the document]
qxz2mqe: fixed

src/provisioning/CatenaX.NetworkServices.Provisioning.Library/Extensions/RoleManager.cs
qxz2mqe: fixed

src/provisioning/CatenaX.NetworkServices.Provisioning.Library/Extensions/RoleManager.cs
```diff
@@ -7,37 +10,54 @@ public partial class ProvisioningManager
         private async Task<(string, IEnumerable<Role>)> GetCentralClientIdRolesAsync(string clientName, IEnumerable<string> roleNames)
```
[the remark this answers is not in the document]
qxz2mqe: fixed

src/provisioning/CatenaX.NetworkServices.Provisioning.Library/Extensions/RoleManager.cs
qxz2mqe: fixed

src/provisioning/CatenaX.NetworkServices.Provisioning.Library/Extensions/RoleManager.cs
qxz2mqe: fixed

src/administration/CatenaX.NetworkServices.Administration.Service/BusinessLogic/UserBusinessLogic.cs
qxz2mqe: fixed

src/administration/CatenaX.NetworkServices.Administration.Service/BusinessLogic/UserBusinessLogic.cs
qxz2mqe: fixed

src/administration/CatenaX.NetworkServices.Administration.Service/BusinessLogic/UserBusinessLogic.cs
qxz2mqe: fixed

qxz2mqe (approved):

PR #107, opened 2022-06-30
CPLP-1120 refactor app business logic

Phil91:
fixing [CPLP-1120](https://jira.catena-x.net/browse/CPLP-1120)

✅ Added Repositories
✅ Access to Repository Factory
✅ App Business Logic accesses DbContext via Repositories (PortalRepository)
✅ fixed and moved unit tests


ntruchsess (approved):
I did some more cleanup of the code, unifying multiple sequentially executed queries into single ones wherever applicable. Fixed activation of subscription that would try to activate a subscription of the provider-company. Adjusted the unit-tests accordingly and run postman successfully against all endpoints.
From my point of view ready to merge. @Phil91 please double-check.

PR #110, opened 2022-07-01
CPLP-1134 Add Notification

Phil91:
Implementation of [CPLP-1130](https://jira.catena-x.net/browse/CPLP-1130), [CPLP-1131](https://jira.catena-x.net/browse/CPLP-1131), [CPLP-1132](https://jira.catena-x.net/browse/CPLP-1132), [CPLP-1133](https://jira.catena-x.net/browse/CPLP-1133) & [CPLP-1134](https://jira.catena-x.net/browse/CPLP-1134)

✅ Add Notification incl. NotificationStatus & NotificationType
✅ Add migration
✅ Add Endpoint to Create, Get, Put Notifications and Get the notification count
✅ Add UnitTests
✅ Add Cors
✅ Edit database scripts

Also see: [PR in devenv](https://github.com/tractusx-team-portal-onboarding/devenv/pull/80)

PR #113, opened 2022-07-04
CPLP-1133 create notification endpoint

Phil91:
✅ Add separate service for notifications
✅ Add endpoint to create notifications
✅ Add UnitTests

Please wait until depending PR #110 is finished and merged into this branch


.github/workflows/trivy.yml
```diff
@@ -172,7 +153,69 @@ jobs:
...
+      # You don't need to set ENV vars when downloading from a public repository.
+      # For public images, no ENV vars must be set.
+      - name: Run Trivy vulnerability scanner
+        if: always()
+        uses: aquasecurity/trivy-action@master
+        with:
+          # Path to Docker image
+          image-ref: "ghcr.io/catenax-ng/product-portal-backend_notification-service:latest"
+          format: "sarif"
+          output: "trivy-results4.sarif"
```
evegufy, line 216: For some reason the diff of this appears a bit strange to me in the Github webv iew, but when I look at the entire file  it seems fine with the exception that the output file should "trivy-results6.sarif" (also in line 223) not 4. Otherwise it's the same as for the app marketplace scan.

evegufy (changes requested):

Phil91:
Changes from this PR will be aggregatet in PR #110

PR #120, opened 2022-07-12
CPLP-1181 add validation to create connector bl

Phil91:
Implementation of [CPLP-1181](https://jira.catena-x.net/browse/CPLP-1181)


Phil91:
@jjeroch I've added a .ToUpper() to the location, which will convert e.g. 'de' to 'DE'. With that some of the errors won't occure. Hope that's fine

src/administration/CatenaX.NetworkServices.Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
```diff
@@ -56,22 +77,46 @@ public Task<Pagination.Response<ConnectorViewModel>> GetAllCompanyConnectorViewM
...
+        {
+            throw new ArgumentException($"Location {connectorInputModel.Location} does not exist", nameof(location));
+        }
+
+        if (!await _portalRepositories.GetInstance<ICompanyRepository>().CheckCompanyExistsByIdAsync(provider).ConfigureAwait(false))
+        {
+            throw new ArgumentException($"Company {connectorInputModel.Provider} does not exist", nameof(provider));
+        }
+
+        if (provider != host && host.HasValue && !await _portalRepositories.GetInstance<ICompanyRepository>().CheckCompanyExistsByIdAsync(host.Value).ConfigureAwait(false))
```
ntruchsess, line 92: If the same repository is used multiple times within a method please get a single reference and reuse it.

ntruchsess (changes requested):

src/administration/CatenaX.NetworkServices.Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
Phil91: fixed it.
just for my understanding, if we don't have the single reference we will use "different databases" correct?

ntruchsess: no we would not have different database(connections), When being called within the same execution-context of the dependency-container those calls are just redundant and besides unnecessary spending a few cycles of cpu-time it would not cause any harm.

src/administration/CatenaX.NetworkServices.Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
```diff
@@ -56,22 +77,47 @@ public Task<Pagination.Response<ConnectorViewModel>> GetAllCompanyConnectorViewM
...
+            HostId = host,
+            TypeId = type,
+            StatusId = status
         };
 
         Connector createdConnector;
         HttpResponseMessage response;
-        using var transaction = await _repository.BeginTransactionAsync();
+
+        await using var transaction = await _portalRepositories.BeginTransactionAsync().ConfigureAwait(false);
```
ntruchsess, line 115: I guess the intention is to run the database-update and the remote api-call atomically within a single transaction? As the api-call does not participate in this transaction it is not any better then calling safeAsync on the dbContext after the api-call and bail out in case the api-call is not succesful.
Please remove the call to saveAsync from the implementation of CreateConnectorAsync() and call it separately after the api-call (is is exposed on the instance of PortalRepositories).

src/marketplace/CatenaX.NetworkServices.App.Service/Controllers/AppsController.cs
```diff
@@ -106,20 +108,21 @@ public AppsController(IAppsBusinessLogic appsBusinessLogic)
     [Authorize(Roles = "add_app")]
     [ProducesResponseType(typeof(Guid), StatusCodes.Status201Created)]
     public async Task<ActionResult<Guid>> CreateAppAsync([FromBody] AppInputModel appInputModel) =>
-        CreatedAtRoute(string.Empty, await this._appsBusinessLogic.CreateAppAsync(appInputModel));
+        CreatedAtRoute(string.Empty, await _appsBusinessLogic.CreateAppAsync(appInputModel).ConfigureAwait(false));
```
ntruchsess, line 111: That is not the propper use of CreateAtRoute. The first parameter shall refer to the name of an endpoint that can be used to retrive this newly created entity. The second paramter is an anynomous object with property-names matching the parameters of that query that is being referred to. The Framework uses this information to create the standard http-response-header 'Location'.
With Route-name being 'string.empty' this ain't work.

src/portalbackend/CatenaX.NetworkServices.PortalBackend.DBAccess/Repositories/AppRepository.cs
```diff
@@ -61,4 +81,114 @@ public async Task<(string appName, string providerName, string providerContactEm
             .Where(appClient => appClient.AppId == appId)
             .Select(appClient => appClient.IamClient!.ClientClientId)
             .SingleOrDefaultAsync();
+
+    /// <inheritdoc />
+    public App CreateApp(Guid id, string provider) =>
```
ntruchsess, line 86: from my point of view it's preferable to create the new Guid within the Create-method.

ntruchsess (changes requested):

src/administration/CatenaX.NetworkServices.Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
Phil91: Didn't recognized that. I've removed the Transaction call and changed it to the SaveAsync if the SdFactory Call was successfull.

src/marketplace/CatenaX.NetworkServices.App.Service/Controllers/AppsController.cs
Phil91: Absolutly right. Changed it

src/portalbackend/CatenaX.NetworkServices.PortalBackend.DBAccess/Repositories/AppRepository.cs
Phil91: The Id is now created in the repository

ntruchsess: I just added a addOptionalParameters Action: https://github.com/catenax-ng/product-portal-backend/pull/120/commits/898ed8d649035916ce575d18397c6c9386958d5c

src/administration/CatenaX.NetworkServices.Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
```diff
@@ -53,25 +75,61 @@ public Task<Pagination.Response<ConnectorViewModel>> GetAllCompanyConnectorViewM
...
+            createdConnector = _portalRepositories.GetInstance<IConnectorsRepository>().CreateConnectorAsync(name, location.ToUpper(), connectorUrl,
+                (connector) =>
+                {
+                    connector.ProviderId = provider;
+                    connector.HostId = host;
+                    connector.TypeId = type;
+                    connector.StatusId = status;
+                });
+            var bpn = (await companyRepository.GetCompanyByIdAsync(connectorInputModel.Provider))!.BusinessPartnerNumber!;
             var httpClient = new HttpClient();
```
ntruchsess, line 133: - the api-call should be encapsulated in its own service-class that gets the required config injected and hides the technical details from the business-logic.
- in line 139 is a catch(Exception) { throw; }. That is redundant and should be removed.

src/administration/CatenaX.NetworkServices.Administration.Service/Controllers/ConnectorsController.cs
```diff
@@ -42,9 +63,18 @@ public ConnectorsController(IConnectorsBusinessLogic connectorsBusinessLogic)
...
+    public Task<Pagination.Response<ConnectorData>> GetCompanyConnectorsForCurrentUserAsync([FromQuery] int page = 0, [FromQuery] int size = 15) =>
+        this.WithIamUserId(iamUserId => _businessLogic.GetAllCompanyConnectorDatasForIamUserAsyncEnum(iamUserId, page, size));
+
+    [HttpGet]
+    [Route("{connectorId}", Name = nameof(GetCompanyConnectorByIdForCurrentUserAsync))]
+    [Authorize(Roles = "view_connectors")]
+    [ProducesResponseType(typeof(ConnectorData), StatusCodes.Status200OK)]
+    [ProducesResponseType(typeof(ErrorResponse), StatusCodes.Status403Forbidden)]
+    [ProducesResponseType(typeof(ErrorResponse), StatusCodes.Status404NotFound)]
+    public Task<ConnectorData> GetCompanyConnectorByIdForCurrentUserAsync([FromRoute] Guid connectorId) =>
```
ntruchsess, line 76: FYI: I did add this getconnectorbyid endpoint. The CreatedAtRoute-response will create a location-header that links the created entity, not the general list of all entities.

src/keycloak/CatenaX.NetworkServices.Keycloak.Authentication/ControllerExtensions.cs
```diff
@@ -24,4 +44,19 @@ public static T WithIamUserId<T>(this ControllerBase controller, Func<string, T>
         }
         return idConsumingFunction(sub);
     }
+
+    public static T WithBearerToken<T>(this ControllerBase controller, Func<string, T> tokenConsumingFunction)
```
ntruchsess, line 48: FYI: I did add this extension-function to the ControllerBase that extracts the bearer-token (same pattern as WithIamUSerId)

src/portalbackend/CatenaX.NetworkServices.PortalBackend.DBAccess/Models/ConnectorData.cs
```diff
@@ -0,0 +1,67 @@
...
+
+using CatenaX.NetworkServices.PortalBackend.PortalEntities.Enums;
+using System.ComponentModel.DataAnnotations;
+
+namespace CatenaX.NetworkServices.PortalBackend.DBAccess.Models;
+
+/// <summary>
+/// View model for connectors.
+/// </summary>
+public class ConnectorData
```
ntruchsess, line 29: FYI: I moved the ViewModel to the DBAccess-Models to allow creating instances within the repository.

src/portalbackend/CatenaX.NetworkServices.PortalBackend.DBAccess/Repositories/ConnectorsRepository.cs
```diff
@@ -21,26 +41,32 @@ public ConnectorsRepository(PortalDbContext portalDbContext)
...
+                    Status = connector.Status!.Id,
+                    Type = connector.Type!.Id
+                },
+                connector.Provider!.CompanyUsers.Any(companyUser => companyUser.IamUser!.UserEntityId == iamUser)
+            ))
+            .SingleOrDefaultAsync();
 
     /// <inheritdoc/>
-    public async Task<Connector> CreateConnectorAsync(Connector connector)
+    public Connector CreateConnectorAsync(string name, string location, string connectorUrl, Action<Connector>? setupOptionalFields)
```
ntruchsess, line 65: Rename to 'CreateConnector' as it is not async

src/portalbackend/CatenaX.NetworkServices.PortalBackend.DBAccess/Repositories/IConnectorsRepository.cs
```diff
@@ -13,24 +33,23 @@ public interface IConnectorsRepository
     /// </summary>
     /// <param name="iamUserId">ID of the iam user used to determine company's connectors for.</param>
     /// <returns>Queryable of connectors that allows transformation.</returns>
-    public IQueryable<Connector> GetAllCompanyConnectorsForIamUser(string iamUserId);
+    IQueryable<Connector> GetAllCompanyConnectorsForIamUser(string iamUserId);
+
+    Task<(ConnectorData ConnectorData, bool IsProviderUser)> GetConnectorByIdForIamUser(Guid connectorId, string iamUser);
```
ntruchsess, line 38: FYI: I did remove the 'public'-keywords as methods of a public interface are public by default. You need to specify 'public' in the implementation

ntruchsess (changes requested):

src/portalbackend/CatenaX.NetworkServices.PortalBackend.DBAccess/Repositories/ConnectorsRepository.cs
Phil91: Renamed

src/administration/CatenaX.NetworkServices.Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
Phil91: i've refactored it to a separate service

ntruchsess (approved):
I did consolidate the multiple company-related queries into a single one and did some cleanup: https://github.com/catenax-ng/product-portal-backend/pull/120/commits/f8eeb0301ddd7c698cc794c92315c06a538deca5
From my point of view it is not ready to merge. Please double-check my changes before merging

PR #132, opened 2022-07-26
feature/CPLP- 828 get all use case

qxz2mqe:
https://jira.catena-x.net/browse/CPLP-828


src/portalbackend/CatenaX.NetworkServices.PortalBackend.DBAccess/Models/UseCaseData.cs
```diff
@@ -0,0 +1,62 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+using System.Text.Json.Serialization;
+
+namespace CatenaX.NetworkServices.PortalBackend.DBAccess.Models;
+
+/// <summary>
+/// model for Use Case.
+/// </summary>
+public class UseCaseData
```
Phil91, line 27: we could use a record here instead of a class, this would look like this:
public record UseCaseData([property: JsonPropertyName("useCaseId")] Guid Id, [property: JsonPropertyName("name")] string Name, [property: JsonPropertyName("shortname")] string ShortName)

but that's just a nice to have in my opinion. @ntruchsess what do you think?

Phil91 (commented):
Maybe a little cosmetical remark, could you add a break at the end of the files? That would result in GitHub not showing the red circle anymore

src/portalbackend/CatenaX.NetworkServices.PortalBackend.DBAccess/Repositories/StaticDataRepository.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// Constructor
+    /// </summary>
+    /// <param name="portalDbContext">PortalDb context.</param>
+    public StaticDataRepository(PortalDbContext portalDbContext)
+    {
+        _dbContext = portalDbContext;
+    }
+
+    /// <inheritdoc />
+    public IAsyncEnumerable<UseCaseData?> GetAllUseCase()
```
Phil91, line 43: Must UseCaseData be nullable? I think we can remove the nullable identifier. If so needs to be changed in the business logic as well

evegufy:
@qxz2mqe Please address the code smell detected by Sonar https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&pullRequest=132&id=catenax-ng_product-portal-backend&open=AYI-dIsLpqte7lcuZaQ6 If necessary please discuss with @ntruchsess how to proceed with the finding. With a failed Sonar scan the PR can't be merged, this applies also to https://github.com/catenax-ng/product-portal-backend/pull/136

src/portalbackend/CatenaX.NetworkServices.PortalBackend.DBAccess/Repositories/StaticDataRepository.cs
ntruchsess: sure, as you cannot have null results but only empty result-set UseCaseData is not nullable here. It has been resolved by @qxz2mqe

ntruchsess (approved):

PR #137, opened 2022-07-27
CPLP-1247 welcome notification

Phil91:
Implementation of [CPLP-1247](https://jira.catena-x.net/browse/CPLP-1247)

✅ Add system generated welcome notifications on application approval 
✅ Extend the notification types
✅ Add a NotificationService to create notifications solution wide
✅ Add UnitTests

Also check the database [PR](https://github.com/tractusx-team-portal-onboarding/devenv/pull/91)


src/CatenaX.NetworkServices.sln
```diff
@@ -86,6 +86,10 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "CatenaX.NetworkServices.Not
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "CatenaX.NetworkServices.Keycloak.ErrorHandling", "keycloak\CatenaX.NetworkServices.Keycloak.ErrorHandling\CatenaX.NetworkServices.Keycloak.ErrorHandling.csproj", "{2FB23833-1283-4A60-9B0C-A6738B09E6ED}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "CatenaX.NetworkServices.Framework.Notifications", "framework\CatenaX.NetworkServices.Framework.Notifications\CatenaX.NetworkServices.Framework.Notifications.csproj", "{70F807A0-B168-4A08-813C-1FB051BABC84}"
```
ntruchsess, line 89: I suggest naming this module CatenaX.NetworkServices.Notification.Library
'FrameWork' should be used for code that is more technical by it's nature so it might also be useful outside the scope of catena and portal.

src/framework/CatenaX.NetworkServices.Framework.Models/Constants.cs
```diff
@@ -34,4 +34,19 @@ public static class Constants
     /// The default language
     /// </summary>
     public const string DefaultLanguage = "en";
+
+    /// <summary>
+    /// Company Name for Catena-X
+    /// </summary>
+    public const string CatenaXCompanyName = "Catena-X";
```
ntruchsess, line 41: those constants do not belong to 'Framework'. They should be configurable in appsettings.json (resp. in the secrets)

src/framework/CatenaX.NetworkServices.Framework.Notifications/NotificationService.cs
```diff
@@ -0,0 +1,104 @@
...
+
+        if (!userIds.Any(x => !x.IsCatenaXAdmin && x.IsCompanyAdmin))
+        {
+            return;
+        }
+
+        foreach (var receiverUserId in userIds.Where(x => !x.IsCatenaXAdmin && x.IsCompanyAdmin).Select(x => x.CompanyUserId))
+        {
+            foreach (var typeId in new[] {
+                         NotificationTypeId.WELCOME,
```
ntruchsess, line 59: this set of NotificationTypeIds should also be configurable in appsettings.json

src/portalbackend/CatenaX.NetworkServices.PortalBackend.DBAccess/Repositories/UserRepository.cs
```diff
@@ -258,6 +258,14 @@ public UserRepository(PortalDbContext portalDbContext)
     /// <inheritdoc />
     public IAsyncEnumerable<(Guid CompanyUserId, bool IsIamUser)> GetCompanyUserWithIamUserCheck(string iamUserId, Guid companyUserId) =>
         _dbContext.CompanyUsers.Where(x => x.IamUser!.UserEntityId == iamUserId || x.Id == companyUserId)
-            .Select(companyUser => ((Guid CompanyUserId, bool IsIamUser)) new (companyUser.Id, companyUser.IamUser!.UserEntityId == iamUserId))
+            .Select(companyUser => ((Guid CompanyUserId, bool IsIamUser)) new ValueTuple<Guid, bool>(companyUser.Id, companyUser.IamUser!.UserEntityId == iamUserId))
+            .ToAsyncEnumerable();
+
+    /// <inheritdoc />
+    public IAsyncEnumerable<(Guid CompanyUserId, bool IsCatenaXAdmin, bool IsCompanyAdmin)> GetCatenaAndCompanyAdminIdAsync(Guid companyId) =>
```
ntruchsess, line 265: repository-methods shall contain business-functionality. In this case the query should not refer to constants, those values should be passed as parameters (where the values in this case should be taken from the appsettings.json by the businesslogic).
It would be even more generic to pass in the roles as IEnumerable and then return the roles that are actually assigned per user also as IEnumerable. The check whether a specific role in this set of roles being queried would then be executed in the business-logic. Of course this will enlarge the size of the result a bit (as role-names are being transfered instead of boolean values) but the method itself then would be 100% free of businesslogic and if we would need to include further roles in the futher the query would not have to change.
There should be no significant difference in execution-time of the query as the very same data needs to be accessed in the db for this.

src/portalbackend/CatenaX.NetworkServices.PortalBackend.PortalEntities/Entities/CompanyUser.cs
```diff
@@ -75,7 +75,7 @@ public CompanyUser(Guid id, Guid companyId, CompanyUserStatusId companyUserStatu
     public virtual ICollection<Document> Documents { get; private set; }
     public virtual ICollection<Invitation> Invitations { get; private set; }
     public virtual ICollection<App> Apps { get; private set; }
-    public virtual ICollection<UserRole> UserRoles { get; private set; }
+    public virtual ICollection<UserRole> UserRoles { get; set; }
     public virtual ICollection<CompanyUserAssignedRole> CompanyUserAssignedRoles { get; private set; }
     public virtual ICollection<CompanyUserAssignedBusinessPartner> CompanyUserAssignedBusinessPartners { get; private set; }
     public virtual ICollection<Notification> Notifications { get; set; }
```
ntruchsess, line 81: same as UserRoles. please change the setter to private and adjust the test-code accordingly.

src/portalbackend/CatenaX.NetworkServices.PortalBackend.PortalEntities/Entities/CompanyUser.cs
```diff
@@ -75,7 +75,7 @@ public CompanyUser(Guid id, Guid companyId, CompanyUserStatusId companyUserStatu
     public virtual ICollection<Document> Documents { get; private set; }
     public virtual ICollection<Invitation> Invitations { get; private set; }
     public virtual ICollection<App> Apps { get; private set; }
-    public virtual ICollection<UserRole> UserRoles { get; private set; }
+    public virtual ICollection<UserRole> UserRoles { get; set; }
```
ntruchsess, line 78: Navigational Properties of x-n relationships shall not be public settable.
In Tests use xxx.UserRoles.Add(...) to seed test-data.

src/notification/CatenaX.NetworkServices.Notification.Service/BusinessLogic/NotificationBusinessLogic.cs
```diff
@@ -30,42 +31,25 @@ namespace CatenaX.NetworkServices.Notification.Service.BusinessLogic;
...
-            {
-                notification.DueDate = dueDate;
-                notification.CreatorUserId = users.Single(x => x.IsIamUser).CompanyUserId;
-                notification.Content = content;
-            });
-
-        await _portalRepositories.SaveAsync().ConfigureAwait(false);
-        return new NotificationDetailData(notification.Id, content, dueDate, notificationTypeId, notificationStatusId);
-    }
+        NotificationCreationData creationData, Guid receiverId) =>
```
ntruchsess, line 49: async + await are redundant, please remove

src/framework/CatenaX.NetworkServices.Framework.Notifications/NotificationService.cs
```diff
@@ -0,0 +1,104 @@
...
+                _portalRepositories.GetInstance<INotificationRepository>().Create(receiverUserId, typeId, false,
+                    notification =>
+                    {
+                        notification.CreatorUserId = userIds.Single(x => x.IsCatenaXAdmin).CompanyUserId;
+                    });
+            }
+        }
+
+
+        await _portalRepositories.SaveAsync().ConfigureAwait(false);
```
ntruchsess, line 75: if this method is meant to be called from other business-logic it should participate in the transaction of that calling code to ensure registration and creation of notifications will happen atomically both or not at all.

src/framework/CatenaX.NetworkServices.Framework.Notifications/NotificationService.cs
```diff
@@ -0,0 +1,104 @@
...
+    ///     Creates a new instance of <see cref="NotificationService" />
+    /// </summary>
+    /// <param name="portalRepositories">Access to the repository factory.</param>
+    public NotificationService(IPortalRepositories portalRepositories)
+    {
+        _portalRepositories = portalRepositories;
+    }
+
+    /// <inheritdoc />
+    public async Task CreateWelcomeNotificationsForCompanyAsync(Guid companyId)
```
ntruchsess, line 43: As this method is called by the Registration-service only it should rather be implemented in that module so configuration of welcome notification type ids would also be part of the registration-settings.
@jjeroch : Is there any intention to send out this specific set of notifications in any other context?

src/framework/CatenaX.NetworkServices.Framework.Notifications/NotificationService.cs
```diff
@@ -0,0 +1,104 @@
...
+            notificationTypeId,
+            notificationStatusId,
+            notification =>
+            {
+                notification.DueDate = dueDate;
+                notification.CreatorUserId = users.Single(x => x.IsIamUser).CompanyUserId;
+                notification.Content = content;
+            });
+
+        await _portalRepositories.SaveAsync().ConfigureAwait(false);
```
ntruchsess, line 101: if this method is meant to be called from other busineslogic (currently this is where the only reference originates from) it should participate in the calling codes transaction.

ntruchsess (changes requested):

src/portalbackend/CatenaX.NetworkServices.PortalBackend.DBAccess/Repositories/UserRepository.cs
```diff
@@ -258,6 +258,14 @@ public UserRepository(PortalDbContext portalDbContext)
     /// <inheritdoc />
     public IAsyncEnumerable<(Guid CompanyUserId, bool IsIamUser)> GetCompanyUserWithIamUserCheck(string iamUserId, Guid companyUserId) =>
         _dbContext.CompanyUsers.Where(x => x.IamUser!.UserEntityId == iamUserId || x.Id == companyUserId)
-            .Select(companyUser => ((Guid CompanyUserId, bool IsIamUser)) new (companyUser.Id, companyUser.IamUser!.UserEntityId == iamUserId))
+            .Select(companyUser => ((Guid CompanyUserId, bool IsIamUser)) new ValueTuple<Guid, bool>(companyUser.Id, companyUser.IamUser!.UserEntityId == iamUserId))
```
ntruchsess, line 261: specifying 'ValueTuple' is not required in dotnet6 any more. It's sufficient to use 'new (companyUser.Id, companyUser.IamUser...)' to create an anonymous tuple-typed object.
(The cast to the specific tuple-type with property-names is still required though).

ntruchsess (changes requested):

src/CatenaX.NetworkServices.sln
Phil91: I've refactored it a bit, guess we don't really need a separat project for it since the notifications are only created in the specific use cases and for general purposes we have our repositories

src/framework/CatenaX.NetworkServices.Framework.Models/Constants.cs
Phil91: i've moved it to the secrets

src/framework/CatenaX.NetworkServices.Framework.Notifications/NotificationService.cs
Phil91: moved as well

src/portalbackend/CatenaX.NetworkServices.PortalBackend.DBAccess/Repositories/UserRepository.cs
Phil91: refactored it.

src/portalbackend/CatenaX.NetworkServices.PortalBackend.PortalEntities/Entities/CompanyUser.cs
Phil91: it's private again

src/portalbackend/CatenaX.NetworkServices.PortalBackend.PortalEntities/Entities/CompanyUser.cs
Phil91: private again :)

src/notification/CatenaX.NetworkServices.Notification.Service/BusinessLogic/NotificationBusinessLogic.cs
Phil91: refactored

src/framework/CatenaX.NetworkServices.Framework.Notifications/NotificationService.cs
Phil91: i refactored it.

src/framework/CatenaX.NetworkServices.Framework.Notifications/NotificationService.cs
Phil91: totaly right, the only use case for the welcome messages is in the bl, refactored it

src/framework/CatenaX.NetworkServices.Framework.Notifications/NotificationService.cs
Phil91: refactored

src/portalbackend/CatenaX.NetworkServices.PortalBackend.DBAccess/Repositories/UserRepository.cs
Phil91: I've added the value tuple due to SonarCloud CodeSmells, i could remove it and we mark it as 'won't fix'

src/portalbackend/CatenaX.NetworkServices.PortalBackend.DBAccess/Repositories/UserRepository.cs
```diff
@@ -258,6 +258,14 @@ public UserRepository(PortalDbContext portalDbContext)
...
+            .ToAsyncEnumerable();
+
+    /// <inheritdoc />
+    public IAsyncEnumerable<(Guid CompanyUserId, bool CompanyIsCatena, IEnumerable<string> RoleNames)> GetCatenaAndCompanyAdminIdAsync(Guid companyId, string catenaXCompanyName, string cxAdminRolename, string companyAdminRole) =>
+        _dbContext.CompanyUsers
+            .Where(x => (x.Company!.Name == catenaXCompanyName && x.UserRoles.Any(y => y.UserRoleText == cxAdminRolename)) ||
```
ntruchsess, line 267: Neither are company.Name and userRoles.UserRoleText primary-key-fields nor are there (and so far also shall not be) any unique constraints that would ensure we actually get the expected result. That is also a security-issue, we must use the companyId and RoleIds here. ( please also negotiate with @jjeroch )

ntruchsess (changes requested):

src/framework/CatenaX.NetworkServices.Framework.Notifications/NotificationService.cs
ntruchsess: cannot see how your refactoring (moving this piece of code into private method in RegistrationBusinessLogic) does address the issue with 'should participate in the same transaction' as the calling code. There are still 2 separate calls to portalrespositories.SaveAsync().

src/administration/CatenaX.NetworkServices.Administration.Service/BusinessLogic/RegistrationBusinessLogic.cs
```diff
@@ -272,6 +252,59 @@ public Task<Pagination.Response<CompanyApplicationWithCompanyUserDetails>> GetAl
...
+
+    private async Task PostRegistrationWelcomeEmailAsync(Guid applicationId)
+    {
+        await foreach (var user in _applicationRepository.GetWelcomeEmailDataUntrackedAsync(applicationId).ConfigureAwait(false))
+        {
+            var userName = String.Join(" ", new[] { user.FirstName, user.LastName }.Where(item => !String.IsNullOrWhiteSpace(item)));
+
+            if (String.IsNullOrWhiteSpace(user.Email))
+            {
+                throw new ArgumentException($"user {userName} has no assigned email");
```
ntruchsess, line 298: I know this is from preexisting code, but if any of the users does not have a email-address assigned the remaining users would not get an email either. (This also results in a code-smell ;-) ).
From my point of view it would be fine to collect the usernames in question (not throwing the exception here but skipping the sending of email) and throwing the ArgumentException once for all users without emailaddress after all other emails have been sent out.
For the very same reason the call to 'SendEmails' should be enclosed in a try-catch as any (even temporary) failure in sending an email (like malformed emal-address) would also stop sending out the remaining emails.

ntruchsess (changes requested):

src/administration/CatenaX.NetworkServices.Administration.Service/BusinessLogic/RegistrationBusinessLogic.cs
```diff
@@ -272,6 +253,67 @@ public Task<Pagination.Response<CompanyApplicationWithCompanyUserDetails>> GetAl
...
+        {
+            return;
+        }
+
+        if (!userIds.Any(x => x.CompanyId == companyId && x.RoleId == _settings.CompanyAdminRoleId))
+        {
+            return;
+        }
+
+        var catenaXAdmin = userIds.First(x => x.CompanyId == _settings.CatenaXCompanyId && x.RoleId == _settings.CxAdminRoleId).CompanyUserId;
```
ntruchsess, line 280: I just realize that this is most likely not what we need here (finding the CX-Admin in the database). Actually the endpoint is called by a CX-Admin user and it would make most sense if this CX-Admin-user calling the endpoint is also the sender of the notifications. @jjeroch : can you confirm this?

ntruchsess (changes requested):

src/administration/CatenaX.NetworkServices.Administration.Service/Controllers/RegistrationController.cs
```diff
@@ -91,15 +91,15 @@ public RegistrationController(IRegistrationBusinessLogic logic)
     /// <response code="500">Internal Server Error.</response>
     /// <response code="502">Bad Gateway Service Error.</response>
     [HttpPut]
-    [Authorize(Roles = "approve_new_partner")]
+    // [Authorize(Roles = "approve_new_partner")]
```
ntruchsess, line 94: why has this role-check been removed?

ntruchsess (changes requested):

src/administration/CatenaX.NetworkServices.Administration.Service/BusinessLogic/RegistrationBusinessLogic.cs
Phil91: i talked with @jjeroch about it. the requesting user should be the creator of the notification. I've already made the changes in the code

src/administration/CatenaX.NetworkServices.Administration.Service/BusinessLogic/RegistrationBusinessLogic.cs
Phil91: I've edited it.

src/administration/CatenaX.NetworkServices.Administration.Service/Controllers/RegistrationController.cs
Phil91: was not supposed to be. it's back in

src/portalbackend/CatenaX.NetworkServices.PortalBackend.DBAccess/Repositories/UserRepository.cs
```diff
@@ -287,6 +287,20 @@ public UserRepository(PortalDbContext portalDbContext)
     /// <inheritdoc />
     public IAsyncEnumerable<(Guid CompanyUserId, bool IsIamUser)> GetCompanyUserWithIamUserCheck(string iamUserId, Guid companyUserId) =>
         _dbContext.CompanyUsers.Where(x => x.IamUser!.UserEntityId == iamUserId || x.Id == companyUserId)
-            .Select(companyUser => ((Guid CompanyUserId, bool IsIamUser)) new (companyUser.Id, companyUser.IamUser!.UserEntityId == iamUserId))
+            .Select(companyUser => new ValueTuple<Guid, bool>(companyUser.Id, companyUser.IamUser!.UserEntityId == iamUserId))
+            .ToAsyncEnumerable();
+
+    /// <inheritdoc />
+    public IAsyncEnumerable<(Guid CompanyUserId, Guid CompanyId, Guid RoleId)> GetCompanyUsersByCompanyAndRoleIdAsync(IEnumerable<(Guid companyId, Guid userRoleId)> companyUserRoleIds) =>
```
ntruchsess, line 294: As it has no references to businesscode but tests any more this method is redundant and may be removed. (Unit-tests need to be adjusted accordingly).

src/administration/CatenaX.NetworkServices.Administration.Service/appsettings.json
```diff
@@ -150,7 +150,12 @@
   "Registration": {
     "ApplicationsMaxPageSize": 20,
     "ApplicationApprovalInitialRoles": {},
-    "PartnerUserInitialRoles": {}
+    "PartnerUserInitialRoles": {},
+    "CatenaXCompanyId": "",
```
ntruchsess, line 154: this is not in sync with the latest code

src/portalbackend/CatenaX.NetworkServices.PortalBackend.DBAccess/Repositories/IUserRepository.cs
```diff
@@ -80,4 +80,11 @@ public interface IUserRepository
     /// <returns><c>true</c> if the user exists, otherwise <c>false</c></returns>
     IAsyncEnumerable<(Guid CompanyUserId, bool IsIamUser)> GetCompanyUserWithIamUserCheck(string iamUserId,
         Guid companyUserId);
+
+    /// <summary>
+    /// Gets the id of the company admin for the given company and the id of the catena x admin
+    /// </summary>
+    /// <param name="companyUserRoleIds">company id with the userRoleId</param>
+    /// <returns>Returns the Id of the user and identifier whether the user is the catena x admin or a company admin</returns>
+    IAsyncEnumerable<(Guid CompanyUserId, Guid CompanyId, Guid RoleId)> GetCompanyUsersByCompanyAndRoleIdAsync(IEnumerable<(Guid companyId, Guid userRoleId)> companyUserRoleIds);
```
ntruchsess, line 89: obsole as of latest code-changes

ntruchsess (changes requested):

ntruchsess (approved):

PR #139, opened 2022-07-27
feat: Build portal-backend helm chart and release workflow

SebastianBezold:
This PR includes a helm chart that is meant to deploy the portal backend.
The default values leave out every Catena-X specifics, so that this chart can be released and used as a dependency on any environment.

Open topics:
- empty appsettings.json leads to pods not starting
- Releasing the chart (workflow only present when on main branch)


evegufy:
Will be merged as part of https://github.com/catenax-ng/product-portal-cd/pull/34

PR #142, opened 2022-07-28
cplp 1151 and 1152 - api(refactoring) response to show the user assigned roles per app

Akshay-Chaudhari-BMW:
refactoring api response to show the user assigned roles per app.


src/administration/CatenaX.NetworkServices.Administration.Service/BusinessLogic/UserBusinessLogic.cs
```diff
@@ -346,7 +346,14 @@ public async Task<CompanyUserDetails> UpdateOwnUserDetails(Guid companyUserId, O
                 companyUser.DateCreated,
                 userData.BusinessPartnerNumbers,
                 companyUser.Company!.Name,
-                companyUser.CompanyUserStatusId)
+                companyUser.CompanyUserStatusId,
+                companyUser.Company!.CompanyAssignedApps
+                    .Where(app => app.AppSubscriptionStatusId == AppSubscriptionStatusId.ACTIVE)
+                    .Select(app => new CompanyUserAssignedRoleDetails(
```
qxz2mqe, line 352: As discussed today Create Record Instead of Class for CompanyUserAssignedRoleDetails

src/administration/CatenaX.NetworkServices.Administration.Service/BusinessLogic/UserBusinessLogic.cs
```diff
@@ -346,7 +346,14 @@ public async Task<CompanyUserDetails> UpdateOwnUserDetails(Guid companyUserId, O
                 companyUser.DateCreated,
                 userData.BusinessPartnerNumbers,
                 companyUser.Company!.Name,
-                companyUser.CompanyUserStatusId)
+                companyUser.CompanyUserStatusId,
+                companyUser.Company!.CompanyAssignedApps
```
ntruchsess, line 350: that data is not being returned by the call to _userRepository.GetUserWithCompanyIdpAsync. Actually below selection-logic should be included in the query to run on the database.
Actually the query currently includes the whole entity Company whereas we only need the companies name and assigned apps. Please take this into account when refactoring the query.

src/administration/CatenaX.NetworkServices.Administration.Service/BusinessLogic/UserBusinessLogic.cs
ntruchsess: done

ntruchsess (changes requested):
please pull once before working on my comment - I did commit the refactoring of CompanyUserAssingedRoleDetails to record (as suggested by @qxz2mqe Majid)

src/administration/CatenaX.NetworkServices.Administration.Service/BusinessLogic/UserBusinessLogic.cs
Akshay-Chaudhari-BMW: resolved

ntruchsess:
The query now returns all roles of all activly subscribed apps regardless of whether they are assigned to that  user or not. So to say the 'assignable roles'.
It rather should return the roles that both are assigned to that user and are roles of apps with active subscription for the users company.

ntruchsess (changes requested):
The query now works but returns other data than being asked for in CPLP-1151 / CPLP-1152

It returns all roles of all activly subscribed apps regardless of whether they are assigned to that user or not. So to say the 'assignable roles'.
It rather should return the roles that both are assigned to that user and are roles of apps with active subscription for the users company.

Akshay-Chaudhari-BMW:
resolved

ntruchsess (approved):

PR #147, opened 2022-08-01
CPLP-1306 Fix CreateConnector

Phil91:


src/portalbackend/CatenaX.NetworkServices.PortalBackend.DBAccess/Repositories/CompanyRepository.cs
```diff
@@ -82,8 +82,9 @@ public CompanyRepository(PortalDbContext portalDbContext)
         _context.Companies
             .AsNoTracking()
             .Where(company => parameters.Select(parameter => parameter.companyId).Contains(company.Id))
-            .Select(company => ((Guid CompanyId, string? BusinessPartnerNumber)) new (
+            .AsAsyncEnumerable()
+            .Select(company => ((Guid CompanyId, string? BusinessPartnerNumber)) new ValueTuple<Guid, string?>(
```
ntruchsess, line 86: What was the error this is trying to fix? Now the select runs on client side and the whole company-entity is selected on the database and being transfered over the network.

ntruchsess (changes requested):

src/administration/CatenaX.NetworkServices.Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
```diff
@@ -95,40 +95,35 @@ public async Task<ConnectorData> CreateConnectorAsync(ConnectorInputModel connec
...
+        var companyDatas = await _portalRepositories
+            .GetInstance<ICompanyRepository>()
+            .GetConnectorCreationCompanyDataAsync(parameters)
             .ToListAsync().ConfigureAwait(false);
 
-        if (!companyDatas.Any(data => data.CompanyId == provider))
+        if (companyDatas.All(data => data.CompanyId != provider))
         {
-            throw new ArgumentException($"Company {provider} does not exist", nameof(provider));
+            throw new UnexpectedConditionException($"Company {provider} does not exist");
```
ntruchsess, line 113: What is the point not to use ArgumentException? UnexpectedConditionException is meant to result in Error 500 but providing ids of not existing entities should result in Error 400 (ids provided as query-parameter or json-body) or Error 404 (ids being part of the request-path).

ntruchsess (changes requested):

src/administration/CatenaX.NetworkServices.Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
Phil91: I've added a ControllerArgumentException, we need to work on that in the future, currently there are 84 references on ArgumentException, i think we need to look at it when ever we touch code that uses it.

src/portalbackend/CatenaX.NetworkServices.PortalBackend.DBAccess/Repositories/CompanyRepository.cs
Phil91: Change it to the select on the database again

src/administration/CatenaX.NetworkServices.Administration.Service/BusinessLogic/ConnectorsBusinessLogic.cs
ntruchsess: I did some more refactoring, changing ConnectorArgumentException to inherit from Exception (not ArgumentException) as it was resulting in the very same code-smell.
(Did also fix the existing smells in GeneralHttpErrorHandler)

ntruchsess (approved):

PR #151, opened 2022-08-02
Feature/cplp 1151 revert cleanup removing redundancy in userdetailselection

ntruchsess:
in PR https://github.com/catenax-ng/product-portal-backend/pull/142 I did fix a code-smell of duplicated lines of code in the UserRepository-class by refactoring those duplicate lines into private methods.
It turned out efcore would not compile those queries the same way. For that reason until we find a working solution to refactor duplicate code in queries that commit needs to be reverted.


Phil91 (approved):
Looks fine by me

PR #160, opened 2022-08-05
CPLP-1351: Add keycloak.net project to cx sln and fix unit test

evegufy:
https://jira.catena-x.net/browse/CPLP-1351


evegufy:
Follow up task for fixing keycloak.net unit tests --> Close PR

PR #164, opened 2022-08-07
Swagger / Endpoint documentation update

jjeroch:
Added summary for "POST/Service Account"


jjeroch:
@Phil91 I have updated a few unfinished endpoint docus.
Unfortunately I havn't been able to correct the following 2 GET endpoints (see screenshot below, the 2 last GET endpoints dont show any "summary" however I could see them in the controller endpoint. Can you please check why they do not show up in swagger?
image

Phil91:
> @Phil91 I have updated a few unfinished endpoint docus. Unfortunately I havn't been able to correct the following 2 GET endpoints (see screenshot below, the 2 last GET endpoints dont show any "summary" however I could see them in the controller endpoint. Can you please check why they do not show up in swagger? image

That was a mean bug... The problem is/was that for example the remark of GetAllCompanyApplicationDetailsAsync contained an unencoded char
Example: GET: api/administration/registration/applicationsWithStatus?page=0&size=15

the '&' char is not html encoded and swagger doesn't render it like that. What we have to do is write the special characters html encoded, for the given scenario it would look like that:

Example: GET: api/administration/registration/applicationsWithStatus?page=0&amp;size=15

as you can see instead of '&' there now is '&amp;' which is the encoded version of &

Phil91 (approved):
Looks fine by me

PR #165, opened 2022-08-08
cplp- 1197 - refactoring api - search app users

Akshay-Chaudhari-BMW:
refactoring api - search all app users


src/portalbackend/CatenaX.NetworkServices.PortalBackend.DBAccess/Repositories/CompanyAssignedAppsRepository.cs
```diff
@@ -44,11 +44,21 @@ public CompanyAssignedAppsRepository(PortalDbContext portalDbContext)
...
+        string iamUserId,
+        string? firstName = null,
+        string? lastName = null,
+        string? email = null,
+        string? roleName = null) =>
+            _context.CompanyUsers
+                .Where(companyUser => companyUser.UserRoles.Any(userRole => userRole.IamClient!.Apps.Any(app => app.Id == appId))
+                 && companyUser.IamUser!.UserEntityId == iamUserId)
+                .SelectMany(companyUser => companyUser.Company!.CompanyUsers)
+                .Where(companyUser => firstName != null ? EF.Functions.ILike(companyUser!.Firstname, $"{firstName}%") : true
```
ntruchsess, line 58: using the string-template is prone to injection if the argument contains any characters that have a special meaning in this context (like firstName containing the character %). Those Characters must either be removed or escaped before building the pattern. Use the 3-parameter version of ILike which allows to specify an escape-paramter and replace all occurences of wildcard-characters (%,_,[,],^) with their escaped version. (Also see: https://docs.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.dbfunctionsextensions.like?view=efcore-6.0 )

ntruchsess (changes requested):

src/portalbackend/CatenaX.NetworkServices.PortalBackend.DBAccess/Repositories/CompanyAssignedAppsRepository.cs
Akshay-Chaudhari-BMW: I used Trim function to escape special characters.

jjeroch:
@Phil91  can you please take over the review finalization

src/portalbackend/CatenaX.NetworkServices.PortalBackend.DBAccess/Repositories/CompanyAssignedAppsRepository.cs
```diff
@@ -44,11 +44,24 @@ public CompanyAssignedAppsRepository(PortalDbContext portalDbContext)
...
-                && companyUser.Company!.CompanyUsers.Any(companyUser => companyUser.IamUser!.UserEntityId == iamUserId));
+    public IQueryable<CompanyUser> GetOwnCompanyAppUsersUntrackedAsync(
+        Guid appId,
+        string iamUserId,
+        string? firstName = null,
+        string? lastName = null,
+        string? email = null,
+        string? roleName = null) {
+
+        char[] MyChar = { '%', '_', '[', ']', '^' };
```
Phil91, line 55: should be myChar

Phil91 (changes requested):
Just a small adjustment

src/portalbackend/CatenaX.NetworkServices.PortalBackend.DBAccess/Repositories/CompanyAssignedAppsRepository.cs
Akshay-Chaudhari-BMW: renamed

Phil91 (approved):

PR #167, opened 2022-08-09
CPLP-1300: Add integration test setup

Phil91:
@evegufy  since Norbert is on vacation i think you are reviewing this pr.
Regarding the two code smells, it's necessary that we have a partial class for the Program, it's needed for our Integration Test setup. But we can't add the static modifier to the Program since the rest of the Program.cs will stop working and we can't build the solution anymore. Same for the namespace code smell. If you'd like we could discuss that in a short call.

Another point i just recognized: we should disable the test coverage check for our test projects 😄


evegufy:
@Phil91 Could you please manage also the 2nd code smell?

Phil91:
> @Phil91 Could you please manage also the 2nd code smell?

I can't resolve the info one somehow

evegufy (approved):
Ok, thanks for the info regarding the code smell.
Looks great!
After you resolved the conflict the PR can be merged.

PR #168, opened 2022-08-11
feature/cplp-1251 company user assigned role db audit

qxz2mqe:
https://jira.catena-x.net/browse/CPLP-1251

PR #169, opened 2022-08-11
feature/cplp-1255 company application db audit

qxz2mqe:
https://jira.catena-x.net/browse/CPLP-1255


Phil91 (commented):
Everything looks good so far. In my opinion we should cancel #168 and use this one. We could then create only one migration for both changes.

Phil91 (approved):
can be merged after [PR](https://github.com/tractusx-team-portal-onboarding/devenv/pull/109) is finished

PR #176, opened 2022-08-17
CPLP-1353 multi tenant app

Phil91:
Resolve Task [CPLP-1353](https://jira.catena-x.net/browse/CPLP-1353)


jjeroch (approved):
Changes and add-ons are reviewed and are looking good.
PR approved

PR #182, opened 2022-08-23
fature/CPLP-827 create app fixed julia comments

qxz2mqe:
https://jira.catena-x.net/browse/CPLP-827


src/marketplace/CatenaX.NetworkServices.App.Service/BusinessLogic/AppsBusinessLogic.cs
```diff
@@ -275,10 +279,17 @@ private async Task<Guid> CreateAppAsync(AppRequestModel appRequestModel)
         appRepository.AddAppLanguages(appRequestModel.SupportedLanguageCodes.Select(c =>
               (appId, c)));
         appRepository.AddAppAssignedUseCases(appRequestModel.UseCaseIds.Select(uc =>
-              (appId, uc)));
+              (appId, Guid.Parse(uc))));
```
Phil91, line 282: what if the conversion fails? for example user sends a request with "useCases": ["justatest" ]

Phil91 (changes requested):

src/marketplace/CatenaX.NetworkServices.App.Service/BusinessLogic/AppsBusinessLogic.cs
qxz2mqe: you are right i didn't thought about it. I will look into this

Phil91 (approved):
looks good

PR #193, opened 2022-09-01
cplp-1246- get own company details.

Akshay-Chaudhari-BMW:
get own company details.


ntruchsess (approved):
did fix the smells and some cleanup

PR #183, opened 2022-08-23
CPLP-1332 sonar

Phil91:
In this PR the tasks [CPLP-1333](https://jira.catena-x.net/browse/CPLP-1333) [CPLP-1334](https://jira.catena-x.net/browse/CPLP-1334) [CPLP-1335](https://jira.catena-x.net/browse/CPLP-1335) are completed

PR #194, opened 2022-09-01
Feature/cplp 591 shared realm service accounts

ntruchsess:
implements https://jira.catena-x.net/browse/CPLP-591
(to avoid conflicts it does contain the changes of #183. When that CR has been merged the number of apparently changed files will be significantly lower, so maybe start the review after #183 has been merged)


ntruchsess:
@evegufy : I did add the master realms role 'admin' to the admin service-account in shared-idp on DEV and INT. That needs to be incorporated into initial setup data for the shared idps.

ntruchsess:
@Phil91 : please also check the corresponding local devenv PR https://github.com/tractusx-team-portal-onboarding/devenv/pull/115

Phil91 (approved):
Looks fine for me.

PR #205, opened 2022-09-10
CPLP-1309: Absolute workdir path and workflow dispatch for KICS

evegufy:
https://jira.catena-x.net/browse/CPLP-1309


docker/Dockerfile-marketplace-app-service
```diff
@@ -7,7 +7,7 @@ FROM mcr.microsoft.com/dotnet/sdk:6.0-alpine AS build
 WORKDIR /
 COPY src/ src/
 COPY keycloak/ keycloak/
-WORKDIR "/src/marketplace/CatenaX.NetworkServices.App.Service"
+WORKDIR /src/marketplace/CatenaX.NetworkServices.App.Service
```
ntruchsess: please change to /src/marketplace/CatenaX.NetworkServices.Apps.Service (Like with 'Services' we did rename 'App' to 'Apps' to avoid name-clashes)

ntruchsess (changes requested):

docker/Dockerfile-marketplace-app-service
evegufy: done

ntruchsess (approved):

PR #209, opened 2022-09-12
CPLP-1413 add subscription details to service details

Phil91:


src/portalbackend/CatenaX.NetworkServices.PortalBackend.DBAccess/Repositories/OfferRepository.cs
```diff
@@ -247,18 +247,19 @@ public async Task<AppDetailsData> GetAppDetailsByIdAsync(Guid appId, string iamU
...
-    public Task<(Guid Id, string? Title, string Provider, string? LeadPictureUri, string? ContactEmail, string? Description, string? Price)> GetServiceDetailByIdUntrackedAsync(Guid serviceId, string languageShortName) => 
+    public Task<ServiceDetailData?> GetServiceDetailByIdUntrackedAsync(Guid serviceId, string languageShortName) => 
         _context.Offers
             .AsNoTracking()
             .Where(x => x.Id == serviceId)
-            .Select(app => new ValueTuple<Guid,string?,string,string?,string?,string?,string?>(
+            .Select(app => new ServiceDetailData(
                 app.Id,
-                app.Name,
+                app.Name?? Constants.ErrorString,
```
ntruchsess, line 256: Why has this been revertet? Setting the ErrorString in the businesslogic is more effizient than doing this on the database-server.

ntruchsess (changes requested):

src/portalbackend/CatenaX.NetworkServices.PortalBackend.DBAccess/Repositories/OfferRepository.cs
Phil91: talked to julia about that, we now return null if the values are empty or not yet set in the database

src/services/CatenaX.NetworkServices.Services.Service/BusinessLogic/ServiceBusinessLogic.cs
```diff
@@ -136,22 +136,15 @@ public async Task<Guid> AddServiceSubscription(Guid serviceId, string iamUserId)
     }
 
     /// <inheritdoc />
-    public async Task<ServiceDetailData> GetServiceDetailsAsync(Guid serviceId, string lang)
+    public async Task<ServiceDetailData> GetServiceDetailsAsync(Guid serviceId, string lang, string iamUserId)
```
ntruchsess, line 139: the Unit-test in ServiceBusinessLogicTest.cs needs to be adjusted accordingly (line 315 ```var result = await sut.GetServiceDetailsAsync(_existingServiceId, "en");```, the call is missing the additional parameter

ntruchsess (changes requested):

ntruchsess (approved):

PR #213, opened 2022-09-14
feature/cplp 830 put app detail

qxz2mqe:
Fix Provided for Error Reported by Julia. RemoveRange Function has some issue. It throw the exception as below. Instead of RemoveRange calling  Remove method which work
Exception thrown: 'System.InvalidOperationException' in Microsoft.EntityFrameworkCore.dll: 'The entity type 'SelectEnumerableIterator<ValueTuple<string, string, string>, OfferDescription>' was not found. Ensure that the entity type has been added to the model.'


src/marketplace/CatenaX.NetworkServices.Apps.Service/Controllers/AppReleaseProcessController.cs
```diff
@@ -49,7 +49,7 @@ public AppReleaseProcessController(IAppReleaseBusinessLogic appReleaseBusinessLo
     }
     
     /// <summary>
-    /// Update active apps in the marketplace for given appId for same company as user
+    /// Update created apps in the marketplace for given appId for same company as user
```
jjeroch, line 52: Change to "Add app details to a newly created owned app under the app release/publishing process."

ntruchsess:
I checked the sources of Select and RemoveRange ( https://source.dot.net/#System.Linq/System/Linq/Select.cs + https://github.com/dotnet/efcore/blob/release/6.0/src/EFCore/DbContext.cs ). RemoveRange is overloaded and it seems the compiler binds to RemoveRange(params: object[]) instead of RemoveRange(IEnumerable<object>). From my understanding there is nothing wrong in our code, Microsoft should have defined it using generigs as RemoveRange<EntityType>(IEnumerable<EntityType>) to resolve the ambiguity.

ntruchsess (changes requested):
I checked the sources of Select and RemoveRange ( https://source.dot.net/#System.Linq/System/Linq/Select.cs + https://github.com/dotnet/efcore/blob/release/6.0/src/EFCore/DbContext.cs ). RemoveRange is overloaded and it seems the compiler binds to RemoveRange(params: object[]) instead of RemoveRange(IEnumerable<object>). From my understanding there is nothing wrong in our code, Microsoft should have defined it using generigs as RemoveRange<EntityType>(IEnumerable<EntityType>) to resolve the ambiguity.

Nevertheless - I'd rather see the foreach-loop within our own (wrapper-) implementation of PortalRepositories.RemoveRange than doing it explicitly in the businesslogic. 

It's also possible that we can force the choice of the correct method by narrowing TEntity to anything derived from class to indicate it's not 'object'.

like:
```
public void RemoveRange<TEntity>(IEnumerable<TEntity> entities) where TEntity : class
        => _dbContext.RemoveRange(entities);
```
(I did add 'where TEntity : class' here). Can you please try wether this fixes the issue? (I checked all compiles fine, but I don't have a testing-scenario ready).

qxz2mqe:
public void RemoveRange<TEntity>(IEnumerable<TEntity> entities) where TEntity : class
        => _dbContext.RemoveRange(entities);

This one has worked. I have committed the changes

ntruchsess (approved):

PR #228, opened 2022-09-21
CPLP-1491 return all roles of a user

Phil91:


ntruchsess (approved):
we forgot to add a where-conditon to the roles. The query would return all roles linked to the offer, not just the ones that are actually assigned to that user. I did fix that. Please double-check

PR #234, opened 2022-09-25
feature/cplp 1343 appdetails with document

qxz2mqe:
https://jira.catena-x.net/browse/CPLP-1343


src/marketplace/Apps.Service/BusinessLogic/AppsBusinessLogic.cs
```diff
@@ -79,9 +79,32 @@ public AppsBusinessLogic(IPortalRepositories portalRepositories, IMailingService
...
+            result.Provider,
+            result.ContactEmail,
+            result.ContactNumber,
+            result.UseCases,
+            result.LongDescription ?? Constants.ErrorString,
+            result.Price ?? Constants.ErrorString,
+            result.Tags,
+            result.IsSubscribed,
+            result.Languages,
+            result.Documents.GroupBy(d => d.documentTypeId).ToDictionary(g => g.Key, g => g.Select(d => new DocumentData(d.documentId, d.documentName)))
```
ntruchsess, line 105: @qxz2mqe : please see this transformation to dictionary using linq instead of foreach-loop.

ntruchsess (approved):
@qxz2mqe as @evegufy did the relocate everything for the eclipse-migration I did merge main. In addition to that I did some refactoring simplifying the logic. Please double-check and test before merging

src/marketplace/Apps.Service/BusinessLogic/AppsBusinessLogic.cs
qxz2mqe: @ntruchsess Tested . It is working fine

PR #243, opened 2022-09-28
CPLP-1473 edc discovery service

qxz2mqe:
https://jira.catena-x.net/browse/CPLP-1473


src/portalbackend/PortalBackend.DBAccess/Repositories/ConnectorsRepository.cs
```diff
@@ -84,4 +84,19 @@ public async Task DeleteConnectorAsync(Guid connectorId)
             throw new NotFoundException("Connector with provided ID does not exist.");
         }
     }
+
+    /// <inheritdoc/>
+    public IAsyncEnumerable<ConnectorCompanyData> GetConnectorEndPointDataAsync(IEnumerable<string> bpns)
+    =>
+        _context.Connectors
+            .AsNoTracking()
+            .Where(connector => connector.Provider!.Id == connector.ProviderId && bpns.Contains(connector.Provider!.BusinessPartnerNumber))
```
ntruchsess, line 93: what shall this condition be good for: 'connector.Provider.Id == connector.ProviderId'
this will be always true as this is how the navigational property is defined.

src/portalbackend/PortalBackend.DBAccess/Repositories/ConnectorsRepository.cs
```diff
@@ -84,4 +84,19 @@ public async Task DeleteConnectorAsync(Guid connectorId)
...
+            .GroupBy(connector => connector.Provider!.BusinessPartnerNumber)
```
ntruchsess, line 94: while it compiles fine 'GroupBy' can be used in ef-queries for aggregates only. (That is because the database does not allow to select columns that are not also included in the group by clause):
see https://www.postgresql.org/docs/14/queries-table-expressions.html#QUERIES-GROUP
Please apply the GroupBy on the query result (in the businesslogic).

ntruchsess (changes requested):

src/administration/Administration.Service/Controllers/ConnectorsController.cs
```diff
@@ -121,4 +121,17 @@ public async Task<IActionResult> DeleteConnectorAsync([FromRoute] Guid connector
...
     }
-}
+
+    /// <summary>
+    /// Retrieves company connector end point for the given bpn.
+    /// </summary>
+    /// <param name="bpns" example="BPNL00000003CRHK">Single or List of Business Partner Number of the company.</param>
+    /// <remarks>Example: GET: /api/administration/connectors/discovery</remarks>
+    /// <response code="200">Returns connector end point along with bpn.</response>
+    [HttpGet]
```
ntruchsess, line 131: FETCH or GET-requests cannot have a body

src/portalbackend/PortalBackend.DBAccess/Repositories/ConnectorsRepository.cs
qxz2mqe: Fixed . I removed GroupBy from Query and applying it in Business Logic

src/portalbackend/PortalBackend.DBAccess/Repositories/ConnectorsRepository.cs
qxz2mqe: Removed

ntruchsess (approved):

PR #245, opened 2022-09-30
CPLP-1379 disable standard flow for service account clients

Phil91:


jjeroch:
@ntruchsess please prioritize the review of this pr - ideally this is included in the release which Evelyn is creating tomorrow. Changes seems to be small, ideally we can do a quick risk based review ;) 
@evegufy  FYI

ntruchsess (approved):

PR #249, opened 2022-10-05
CPLP-1227 bulk-upload own company shared idp users

ntruchsess:
implements https://jira.catena-x.net/browse/CPLP-1227


src/framework/Framework.IO/CsvParser.cs
```diff
@@ -0,0 +1,143 @@
...
+            {
+                throw new ControllerArgumentException($"value for {itemName} type string expected", documentParameterName ?? DefaultDocumentParameterName);
+            }
+            yield return items.Current;
+        }
+    }
+
+    public static async ValueTask<(int Processed, int Lines, IEnumerable<(int Line, Exception Error)> Errors)> ProcessCsvAsync<TLineType>(
+        Stream stream,
+        Action<string> validateFirstLine,
```
Phil91, line 86: I'm thinking about the implementation of the first line validation, we could also hand over the HeaderColumns and validate those, than we don't need to execute an action from outside. But I'm not sure about that. The positive effect of passing the action from outside is that we could validate all kind of stuff (even service calls).

Phil91 (changes requested):
Please provide unit tests, in my opinion espacially the csv parser is perfect to test with specific csv files for each case we could add into the test.

ntruchsess:
I did refactor the existing file-upload that is used in identityprovider user-migration to use the new framework implementation, so no csv-upload with individually implementation of the algorithm remains. Please check.

Phil91 (commented):
changes in identityprovider looks fine, as everything else. we have 6 code smells, we should look into.
About the code duplication, I'm not sure why's that. Can you tell what sonars problem is?

ntruchsess:
All smells are fixed. To make sonar happy I had to apply small refactoring to the usercreation to reduce the complexity.
I did refactor all usecases that create companyUsers to make use of the UserCreationService-implementation. In addition to that I fixed the open 'TODO' in userdeletion not covering own-ipd or mixed idp scenarios.
Can you please re-check?

ntruchsess:
regarding the duplication: I don't think sonar has a problem. There is some duplication (the stream implementations) that technically cannot be avoided as one has to implement the whole interface.

Phil91 (commented):
Code changes look fine to me. We should at least add UnitTests for the Upload

ntruchsess:
@Phil91 guess test-coverage is now sufficient ;-). Can you please do a final re-review?

Phil91 (changes requested):
Review done. Looks good so far. Tested it - works fine.

Locally one Test is running into an error - TestAsyncCopyToSuccess in AsyncEnumerableStringStreamTest, is it only on my machine or do you get the error as well?

If its on my machine I'm going to approve it

Phil91 (approved):
Looks really good! 👍

PR #252, opened 2022-10-06
CPLP-1491 user endpoint

Phil91:


src/portalbackend/PortalBackend.DBAccess/Repositories/UserRepository.cs
```diff
@@ -396,4 +396,29 @@ public UserRepository(PortalDbContext portalDbContext)
...
+        string? firstName = null,
+        string? lastName = null,
+        string? email = null,
+        string? roleName = null) 
+    {
+        char[] escapeChar = { '%', '_', '[', ']', '^' };
+        return _dbContext.CompanyUsers.AsNoTracking()
+            .Where(companyUser => companyUser.UserRoles.Any(userRole => userRole.Offer!.Id == appId) && companyUser.IamUser!.UserEntityId == iamUserId)
+            .SelectMany(companyUser => companyUser.Company!.CompanyUsers)
+            .Where(companyUser => firstName == null || EF.Functions.ILike(companyUser.Firstname!, $"%{firstName.Trim(escapeChar)}%")
```
ntruchsess, line 419: as this not real 'escaping' but just trimming the special characters, did you try searching with % or _ within the search string checking whether you get an exact match with names that actually do include % or _ while not matching the % or _ as a wildcard?
see https://www.postgresql.org/docs/current/functions-matching.html how postgres does treat those values.

ntruchsess (changes requested):

src/portalbackend/PortalBackend.DBAccess/Repositories/UserRepository.cs
Phil91: i just moved the code, i will check it.
but should it really be possible to have company users with an % or _ in the firstname?
at least the % could open the possibility to mess with our database.

Phil91: escaping added

src/portalbackend/PortalBackend.DBAccess/Repositories/UserRepository.cs
```diff
@@ -420,15 +421,19 @@ public UserRepository(PortalDbContext portalDbContext)
...
+        email = email == null ? null : regex.Replace(email, $"%{regex.Replace(email!, @"\")}%"); 
+        roleName = roleName == null ? null : regex.Replace(roleName, $"%{regex.Replace(roleName!, @"\")}%"); 
         return _dbContext.CompanyUsers.AsNoTracking()
             .Where(companyUser => companyUser.UserRoles.Any(userRole => userRole.Offer!.Id == appId) && companyUser.IamUser!.UserEntityId == iamUserId)
             .SelectMany(companyUser => companyUser.Company!.CompanyUsers)
-            .Where(companyUser => firstName == null || EF.Functions.ILike(companyUser.Firstname!, $"%{firstName.Trim(escapeChar)}%")
-                && lastName == null || EF.Functions.ILike(companyUser.Lastname!, $"%{lastName!.Trim(escapeChar)}%")
-                && email == null || EF.Functions.ILike(companyUser.Email!, $"%{email!.Trim(escapeChar)}%")
-                && roleName == null || companyUser.UserRoles.Any(userRole => EF.Functions.ILike(userRole.UserRoleText, $"{roleName!.Trim(escapeChar)}%")));
+            .Where(companyUser => firstName == null || EF.Functions.ILike(companyUser.Firstname!, $"%{regex.Replace(firstName!, @"\")}%")
```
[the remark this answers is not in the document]
ntruchsess: I did fix this.

ntruchsess (approved):
I did fix the operator-precedence issue and removed the characters that (according to https://www.postgresql.org/docs/7.3/functions-matching.html) are not special characters in this context. Please re-check before merging.

PR #260, opened 2022-10-11
CPLP-1470 refactor owncompany search

qxz2mqe:
https://jira.catena-x.net/browse/CPLP-1470

### DIFF
--- a/src/portalbackend/PortalBackend.DBAccess/Repositories/UserRepository.cs
+++ b/src/portalbackend/PortalBackend.DBAccess/Repositories/UserRepository.cs
@@ -123,7 +123,7 @@ public class UserRepository : IUserRepository
                 && companyUserId.HasValue ? companyUser.Id == companyUserId!.Value : true
                 && firstName != null ? companyUser.Firstname == firstName : true
                 && lastName != null ? companyUser.Lastname == lastName : true
-                && email != null && EF.Functions.ILike(companyUser.Email!, $"%{email.Trim(escapeChar)}%"));
+                && email != null ? EF.Functions.ILike(companyUser.Email!, $"%{email.Trim(escapeChar)}%") : true);
         }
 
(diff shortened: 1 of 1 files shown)


src/portalbackend/PortalBackend.DBAccess/Repositories/UserRepository.cs
```diff
@@ -123,7 +123,7 @@ public UserRepository(PortalDbContext portalDbContext)
                 && companyUserId.HasValue ? companyUser.Id == companyUserId!.Value : true
                 && firstName != null ? companyUser.Firstname == firstName : true
                 && lastName != null ? companyUser.Lastname == lastName : true
-                && email != null && EF.Functions.ILike(companyUser.Email!, $"%{email.Trim(escapeChar)}%"));
+                && email != null ? EF.Functions.ILike(companyUser.Email!, $"%{email.Trim(escapeChar)}%") : true);
```
ntruchsess, line 126: sonar sees this as a code-smell. I suggest:
```&& (email == null || EF.Functions.ILike(companyUser.Email!, $"%{email.Trim(escapeChar)}%"))```
(please adjust the other ternary-operators accordingly)

ntruchsess (changes requested):

src/portalbackend/PortalBackend.DBAccess/Repositories/UserRepository.cs
qxz2mqe: fixed

ntruchsess: as && has higher precedence than || the  ||-part must be enclosed in parantheses, otherwise the && part groups with the preceeding condition

qxz2mqe: fixed

ntruchsess (approved):

Phil91 (approved):

PR #263, opened 2022-10-12
CPLP-1653 document endpoint

Phil91:
remove documentId as a query parameter


Phil91:
Had a rebase going wrong. will create a new pr

PR #270, opened 2022-10-14
CPLP-1599 delete appuser role

qxz2mqe:
https://jira.catena-x.net/browse/CPLP-1599


src/portalbackend/PortalBackend.DBAccess/Repositories/OfferRepository.cs
```diff
@@ -305,4 +306,31 @@ public Offer CreateOffer(string provider, OfferTypeId offerType, Action<Offer>?
...
+                offer.UserRoles.Any(userRole => userRole.Id == roleId)
+            ))
+            .SingleOrDefaultAsync();
+
+    ///<inheritdoc/>
+    public async Task DeleteAppRoleAsync(Guid appId, Guid roleId)
+    {
+        try
+        {
+            var userRole = new UserRole(roleId, string.Empty, appId);
```
ntruchsess, line 326: only the primary key is required to remove the entity. The appId is not part of the primary key, only the id is.
Initialize the to be deleted entity as ```new UserRole(roleId,null!,null!);```

src/portalbackend/PortalBackend.DBAccess/Repositories/OfferRepository.cs
```diff
@@ -305,4 +306,31 @@ public Offer CreateOffer(string provider, OfferTypeId offerType, Action<Offer>?
...
+            _context.UserRoles.Attach(userRole);
```
ntruchsess, line 327: calling attach before remove is redundant.

src/portalbackend/PortalBackend.DBAccess/Repositories/OfferRepository.cs
```diff
@@ -305,4 +306,31 @@ public Offer CreateOffer(string provider, OfferTypeId offerType, Action<Offer>?
...
+            _context.UserRoles.Remove(userRole);
+            await _context.SaveChangesAsync().ConfigureAwait(false);
```
ntruchsess, line 329: for consistency with other code the call to SaveChangesAsync shall be done in the businesslogic.

ntruchsess (changes requested):

src/portalbackend/PortalBackend.DBAccess/Repositories/OfferRepository.cs
qxz2mqe: fixed

src/portalbackend/PortalBackend.DBAccess/Repositories/OfferRepository.cs
qxz2mqe: fixed

ntruchsess (approved):

PR #271, opened 2022-10-14
CPLP-1644 fix user email field

Phil91:


src/marketplace/Offers.Library/Service/OfferSetupService.cs
```diff
@@ -46,12 +47,13 @@ public async Task AutoSetupOffer(OfferThirdPartyAutoSetupData autoSetupData, str
 
         try
         {
-            _logger.LogInformation("OfferSetupService was called with the following url: {serviceDetailsAutoSetupUrl}", serviceDetailsAutoSetupUrl);
+            // TODO: Remove the autosetupdata from logging after testing.
+            _logger.LogInformation("OfferSetupService was called with the following url: {ServiceDetailsAutoSetupUrl} and following data: {AutoSetupData}", serviceDetailsAutoSetupUrl, JsonConvert.SerializeObject(autoSetupData));
```
ntruchsess, line 51: please System.Text.Json instead of NewtonSoft https://learn.microsoft.com/de-de/dotnet/standard/serialization/system-text-json/how-to?pivots=dotnet-6-0

ntruchsess (changes requested):

src/marketplace/Offers.Library/Service/OfferSetupService.cs
Phil91: changed to System.Text.Json

ntruchsess (approved):

PR #279, opened 2022-10-19
CPLP-1685 connector host

Phil91:
In this PR [CPLP-1682](https://jira.catena-x.net/browse/CPLP-1682) & [CPLP-1685](https://jira.catena-x.net/browse/CPLP-1685) are resolved


ntruchsess (approved):
did some cleanup, please double-check before merging

PR #285, opened 2022-10-24
CPLP-734-exclude-delete-status

qxz2mqe:
https://jira.catena-x.net/browse/CPLP-734

### DIFF
--- a/src/administration/Administration.Service/BusinessLogic/UserSettings.cs
+++ b/src/administration/Administration.Service/BusinessLogic/UserSettings.cs
@@ -19,6 +19,7 @@
  ********************************************************************************/
 
 using System.ComponentModel.DataAnnotations;
+using Org.CatenaX.Ng.Portal.Backend.PortalBackend.PortalEntities.Enums;
 
 namespace Org.CatenaX.Ng.Portal.Backend.Administration.Service.BusinessLogic;
 
@@ -34,6 +35,13 @@ public class UserSettings
     public UserSetting Portal { get; set; }
     public PasswordReset PasswordReset { get; set; }
     public int ApplicationsMaxPageSize { get; set; }
+
+    /// <summary>
+    /// Document Type Id
+    /// </summary>
+    /// <value></value>
+    [Required]
+    public IEnumerable<CompanyUserStatusId> CompanyUserStatusIds { get; set; } = null!;
 }
 
 public class UserSetting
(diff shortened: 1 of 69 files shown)


src/administration/Administration.Service/BusinessLogic/UserSettings.cs
```diff
@@ -34,6 +35,13 @@ public UserSettings()
     public UserSetting Portal { get; set; }
     public PasswordReset PasswordReset { get; set; }
     public int ApplicationsMaxPageSize { get; set; }
+
+    /// <summary>
+    /// Document Type Id
```
jjeroch, line 40: @qxz2mqe why do we add something related to "Document Type Id"?

tests/administration/Administration.Service.Tests/BusinessLogic/UserBusinessLogicTests.cs
```diff
@@ -99,8 +99,13 @@ public UserBusinessLogicTests()
...
         _processLine = A.Fake<Func<UserCreationInfoIdp,(Guid CompanyUserId, string UserName, string? Password, Exception? Error)>>();
         _companyUserSelectFunction = A.Fake<Func<CompanyUser,CompanyUser>>();
-
+        _settings = A.Fake<UserSettings>();
         _error = _fixture.Create<TestException>();
+        _settings.CompanyUserStatusIds = new List<CompanyUserStatusId>
+        {
+            CompanyUserStatusId.ACTIVE,
+            CompanyUserStatusId.INACTIVE
+        };
```
jjeroch, line 108: @qxz2mqe  the test validates that active and inactive user status is received. Does it also validate inside the example that any other status especially DELETE is not included in the response?

ntruchsess: it might be reasonable not to configure the CompanyUserStatus that are taken as active but to configure the status that is to be ignored. In case we add further status to the enum they would be ignored in case this configuration is not being adjusted accordingly.

ntruchsess (changes requested):
https://jira.catena-x.net/browse/CPLP-734 is about adjusting 'GET /administration/user/owncompany/users'. You did modify 'DELETE /administration/user/owncompany/users'

jjeroch:
@qxz2mqe if all the comments added above are handled in the new pr - please delete this old pr and branch

qxz2mqe:
@jjeroch  I will delete the old PR

PR #292, opened 2022-10-26
CPLP-1540 multi single app

Phil91:


Phil91:
Closed due to wrong push. New PR is created #293

PR #296, opened 2022-10-28
CPLP-734-getowncompanyuser-excludedeletestatus

qxz2mqe:
https://jira.catena-x.net/browse/CPLP-734


src/administration/Administration.Service/BusinessLogic/UserBusinessLogic.cs
```diff
@@ -247,6 +247,7 @@ public async Task<Guid> CreateOwnCompanyIdpUserAsync(Guid identityProviderId, Us
                 companyUsers.OrderByDescending(companyUser => companyUser.DateCreated)
                 .Skip(skip)
                 .Take(take)
+                .Where(companyUser =>_settings.CompanyUserStatusIds.Contains(companyUser.CompanyUserStatusId))
```
ntruchsess, line 250: that needs to be relocated into the base-query. Otherwise the pagination-metadata (count etc..) would not match the current page content. Also applying the Where after Skip() and Take() would reduce the number of items being returned on this particular page.

ntruchsess (changes requested):

jjeroch (dismissed):
@qxz2mqe please also check and comment why the unit tests are failing. Can we get them running successfully?

src/administration/Administration.Service/BusinessLogic/UserBusinessLogic.cs
qxz2mqe: I would like to discuss with you because same query method is being used for another end point

qxz2mqe:
@jjeroch  It was failing for other PR which I fixed

src/administration/Administration.Service/BusinessLogic/UserBusinessLogic.cs
ntruchsess: did fix my review-comment: https://github.com/catenax-ng/product-portal-backend/pull/296/commits/768e5a6c411784ff0c00a39baf3e2df0b7356ab8

ntruchsess (approved):

PR #304, opened 2022-11-03
CPLP-1501 add filter for has role

Phil91:
In this PR [CPLP-1501](https://jira.catena-x.net/browse/CPLP-1501) is solved


src/portalbackend/PortalBackend.DBAccess/Repositories/UserRepository.cs
```diff
@@ -412,8 +413,9 @@ public CompanyUser AttachAndModifyCompanyUser(Guid companyUserId, Action<Company
         roleName = roleName == null ? null : regex.Replace(roleName!, @"\");
 
         return _dbContext.CompanyUsers.AsNoTracking()
-            .Where(companyUser => companyUser.UserRoles.Any(userRole => userRole.Offer!.Id == appId) &&
-                                  companyUser.IamUser!.UserEntityId == iamUserId)
+            .Where(companyUser => companyUser.IamUser!.UserEntityId == iamUserId && 
+                                  (!hasRole.HasValue || !hasRole.Value || companyUser.UserRoles.Any(userRole => userRole.Offer!.Id == appId)) &&
```
ntruchsess, line 417: according to https://jira.catena-x.net/browse/CPLP-1501 that filter shall apply to the users that belong to the same company as the user calling the endpoint.
Here the criteria is applied to the user who is calling the endpoint.
That means:
- if hasRole is true but the calling user does not have a role of this offer assigned the result will be empty. 
- if hasRole is false but the calling user does have a role of this offer assigned the result will be empty.
- if hasRole is null the result depends on the other filter-criteria
The 'hasRole'-condition must be moved to the second where-condition (which belongs to the SelectMany).

ntruchsess (changes requested):

src/portalbackend/PortalBackend.DBAccess/Repositories/UserRepository.cs
Phil91: thanks for recognizing. i've overread it - fixed.

src/portalbackend/PortalBackend.DBAccess/Repositories/UserRepository.cs
```diff
@@ -414,14 +415,15 @@ public CompanyUser AttachAndModifyCompanyUser(Guid companyUserId, Action<Company
         roleName = roleName == null ? null : regex.Replace(roleName!, @"\");
 
         return _dbContext.CompanyUsers.AsNoTracking()
-            .Where(companyUser => companyUser.UserRoles.Any(userRole => userRole.Offer!.Id == appId) &&
```
ntruchsess, line 417: that line made sure the calling user does have at least one of the roles of this offer assigned.
Without the endpoint allows to query any appid.
Nevertheless - the check should verify the company has a subscription for this app as while being an administrator the person calling this endpoint might not (yet) have any roles of this app assigned.
I suggest to include in the where-condition:
```companyUser.Company!.OfferSubscriptions.Any(subscription => subscription.OfferId == appId && statusIds.Contains(subscription.OfferSubscriptionStatusId)))```
passing ```IEnumerable<OfferSubscriptionStatusId> statusIds``` to the query so the businesslogic would define which Subscription-status would be included in the result.

**Review by ntruchsess — CHANGES_REQUESTED** [review]

File: src/administration/Administration.Service/BusinessLogic/UserBusinessLogic.cs [hunk 3]
```diff
@@ -562,7 +559,7 @@ public async Task<int> DeleteOwnUserBusinessPartnerNumbersAsync(Guid companyUser
 
         if (userWithBpn.UserEntityId == null)
         {
-            throw new Exception($"user {companyUserId} is not associated with a user in keycloak");
+            throw new ArgumentException($"user {companyUserId} is not associated with a user in keycloak");
```
ntruchsess, line 562: use ConflictException as this condition indicates that while the companyUserId itself is valid the users record in the database is incomplete as it is not linked to a keycloak user.

src/administration/Administration.Service/Models/CompanyAppUsersFilter.cs
```diff
@@ -0,0 +1,3 @@
+﻿namespace Org.CatenaX.Ng.Portal.Backend.Administration.Service.Models;
+
+public record CompanyAppUsersFilter(int Page, int Size, string? FirstName, string? LastName, string? Email, string? RoleName, bool? HasRole);
```
ntruchsess, line 3: add newline at the end of the file :-)

ntruchsess (changes requested):

src/administration/Administration.Service/BusinessLogic/UserBusinessLogic.cs
Phil91: done

src/administration/Administration.Service/Models/CompanyAppUsersFilter.cs
Phil91: done :) header included as well

src/portalbackend/PortalBackend.DBAccess/Repositories/UserRepository.cs
Phil91: i've added the statusId condition.

ntruchsess (approved):

PR #309, opened 2022-11-04
CPLP-Fix-Registration-Get-Document-Auth-Role

jjeroch:
Hi @Phil91 I just need to fix the permission name for the endpoint. Is this enough or do we need changes on other areas as well?


Phil91 (approved):
Looks fine

Phil91:
> Hi @Phil91 I just need to fix the permission name for the endpoint. Is this enough or do we need changes on other areas as well?

@jjeroch if not already existing, we need to add or update the value in keycloak as well. but codewise that's the only change we need to do :)

jjeroch:
devenv update running, dev and int templates are already up to date with the new template files

PR #311, opened 2022-11-08
CPLP-1744 handle multiple subscriptions or app-instances when editing role-assignements

ntruchsess:
implements fix for https://jira.catena-x.net/browse/CPLP-1744


Phil91 (approved):
Looks fine.

PR #326, opened 2022-11-16
CPLP-1601-app-roles-activeapp

qxz2mqe:
https://jira.catena-x.net/browse/CPLP-1601


ntruchsess (approved):
@qxz2mqe : I did some cleanup removing the redundancy with preexisting code, please pull in case you need to add further changes.
I noticed Offer.Name is nullable in the db. That implies the notification might not contain a valid app name (if it is not yet set). Please clarify with @jjeroch whether this is ok or should result in an error indicating the app-name must be set before.

ntruchsess (changes requested):
please add unit-tests for the new code

qxz2mqe:
@ntruchsess Added Unit Test

ntruchsess (approved):

PR #333, opened 2022-11-17
CPLP-914-post-company-bpn

qxz2mqe:
https://jira.catena-x.net/browse/CPLP-914

### DIFF
--- a/src/administration/Administration.Service/BusinessLogic/RegistrationBusinessLogic.cs
+++ b/src/administration/Administration.Service/BusinessLogic/RegistrationBusinessLogic.cs
@@ -358,42 +358,43 @@ public class RegistrationBusinessLogic : IRegistrationBusinessLogic
         return roleData;
     }
 
-    public Task<bool> UpdateCompanyBpn(Guid applicationId, string bpn, string iamuserId)
+    public Task UpdateCompanyBpn(Guid applicationId, string bpn, string iamuserId)
     {
         var regex = new Regex(@"(\w|\d){16}");
         if (!regex.IsMatch(bpn))
         {
-            throw new ArgumentException("BPN must contain exactly 16 characters long.", nameof(bpn));
+            throw new ControllerArgumentException("BPN must contain exactly 16 characters long.", nameof(bpn));
         }
-        if (!bpn.ToUpper().StartsWith("BPNL"))
+        if (!bpn.ToUpper().StartsWith("BPNL", StringComparison.OrdinalIgnoreCase))
         {
-            throw new ArgumentException("businessPartnerNumbers must prefixed with BPNL", nameof(bpn));
+            throw new ControllerArgumentException("businessPartnerNumbers must prefixed with BPNL", nameof(bpn));
         }
         return UpdateCompanyBpnAsync(applicationId, bpn, iamuserId);
     }
 
-    private async Task<bool> UpdateCompanyBpnAsync(Guid applicationId, string bpn, string iamuserId)
+    private async Task UpdateCompanyBpnAsync(Guid applicationId, string bpn, string iamuserId)
     {
-        var businessPartnerNumbers = await _portalRepositories.GetInstance<ICompanyRepository>().GetAllMemberCompaniesBPNAsync().ToListAsync().ConfigureAwait(false);
-        if (businessPartnerNumbers.Contains(bpn))
+        await foreach (var item in _portalRepositories.GetInstance<IUserRepository>().GetBpnForIamUserUntrackedAsync(iamuserId, applicationId, bpn).ConfigureAwait(false))
         {
-            throw new ControllerArgumentException($"BusinessPartnerNumber (bpn) is already exist", nameof(bpn));
-        }
-        var result = await _portalRepositories.GetInstance<IUserRepository>().GetBpnForIamUserUntrackedAsync(iamuserId, applicationId).ConfigureAwait(false);
-        if (result == default)
-        {
-            throw new NotFoundException($"user {iamuserId} is not associated with any company");
-        }
-        if (string.IsNullOrWhiteSpace(result.Bpn))
-        {
-            _portalRepositories.GetInstance<ICompanyRepository>().AttachAndModifyCompany(result.CompanyId, c =>
+            if (item == default)
             {
-                c.BusinessPartnerNumber = bpn;
-            });
+                throw new NotFoundException($"user {iamuserId} is not associated with any company");
+            }
+            if (!item.IsApplicationCompany)
+            {
+                throw new ConflictException($"Bpn can not be updated for applicationId");
+            }
+            if (string.IsNullOrWhiteSpace(item.BusinessPartnerNumber))
+            {
+                _portalRepositories.GetInstance<ICompanyRepository>().AttachAndModifyCompany(item.CompanyId, c =>
+                {
+                    c.BusinessPartnerNumber = bpn;
+                });
 
-            await _portalRepositories.SaveAsync().ConfigureAwait(false);
-            return true;
+                await _portalRepositories.SaveAsync().ConfigureAwait(false);
+            }
+
         }
-        return false;
+        
     }
 }
--- a/src/administration/Administration.Service/BusinessLogic/RegistrationBusinessLogic.cs
+++ b/src/administration/Administration.Service/BusinessLogic/RegistrationBusinessLogic.cs
@@ -358,7 +358,7 @@ public class RegistrationBusinessLogic : IRegistrationBusinessLogic
         return roleData;
     }
 
-    public async Task<bool> UpdateCompanyBpn(Guid applicationId, string bpn, string iamuserId)
+    public Task<bool> UpdateCompanyBpn(Guid applicationId, string bpn, string iamuserId)
     {
         var regex = new Regex(@"(\w|\d){16}");
         if (!regex.IsMatch(bpn))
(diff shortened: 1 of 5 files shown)


src/administration/Administration.Service/BusinessLogic/RegistrationBusinessLogic.cs
```diff
@@ -356,4 +357,43 @@ private static async Task<List<UserRoleData>> GetRoleData(IUserRolesRepository u
...
+        {
+            throw new ArgumentException("businessPartnerNumbers must prefixed with BPNL", nameof(bpn));
+        }
+        return UpdateCompanyBpnAsync(applicationId, bpn, iamuserId);
+    }
+
+    private async Task<bool> UpdateCompanyBpnAsync(Guid applicationId, string bpn, string iamuserId)
+    {
+        var businessPartnerNumbers = await _portalRepositories.GetInstance<ICompanyRepository>().GetAllMemberCompaniesBPNAsync().ToListAsync().ConfigureAwait(false);
+        if (businessPartnerNumbers.Contains(bpn))
```
ntruchsess, line 378: reading all datasets to check the existence of a specific one on client-side is inefficiency at its worst. Please integrate this check into the query being called in line 382 (GetBpnForIamUserUntrackedAsync)

src/administration/Administration.Service/BusinessLogic/RegistrationBusinessLogic.cs
```diff
@@ -356,4 +357,43 @@ private static async Task<List<UserRoleData>> GetRoleData(IUserRolesRepository u
...
     }
+
+    public Task<bool> UpdateCompanyBpn(Guid applicationId, string bpn, string iamuserId)
+    {
+        var regex = new Regex(@"(\w|\d){16}");
+        if (!regex.IsMatch(bpn))
+        {
+            throw new ArgumentException("BPN must contain exactly 16 characters long.", nameof(bpn));
+        }
+        if (!bpn.ToUpper().StartsWith("BPNL"))
```
ntruchsess, line 368: please use overload StartsWith(String, StringComparison) that allows to do the comparism ignoring case in a single step. (see: https://learn.microsoft.com/en-us/dotnet/api/system.string.startswith?view=net-6.0#system-string-startswith(system-string-system-stringcomparison) )

src/administration/Administration.Service/BusinessLogic/RegistrationBusinessLogic.cs
```diff
@@ -356,4 +357,43 @@ private static async Task<List<UserRoleData>> GetRoleData(IUserRolesRepository u
...
+        if (businessPartnerNumbers.Contains(bpn))
+        {
+            throw new ControllerArgumentException($"BusinessPartnerNumber (bpn) is already exist", nameof(bpn));
+        }
+        var result = await _portalRepositories.GetInstance<IUserRepository>().GetBpnForIamUserUntrackedAsync(iamuserId, applicationId).ConfigureAwait(false);
+        if (result == default)
+        {
+            throw new NotFoundException($"user {iamuserId} is not associated with any company");
+        }
+        if (string.IsNullOrWhiteSpace(result.Bpn))
```
ntruchsess, line 387: this is not implementing the requirement 'bpn can only get added, not changed (means if the company already has an bpn, the admin can not change the bpn; BUT the admin can add an bpn if there is no bpn yet added)' that is specified in https://jira.catena-x.net/browse/CPLP-914
obviously to add an additional BusinessPartnerNumber to a company that already does have a BusinessPartnerNumber assigned we'd need to change the database-scheme to allow 1-n releationship in between company and BusinessPartnerNumber adding a separate BusinessPartnerNumber-entity that links back to the company. We need to discuss with @jjeroch how to proceed here

ntruchsess (changes requested):

src/administration/Administration.Service/BusinessLogic/RegistrationBusinessLogic.cs
ntruchsess, line 387: in case the BusinessPartnerNumber is already set I suggest to throw ConflictException to indicate the update cannot be done.

ntruchsess (changes requested):

src/administration/Administration.Service/BusinessLogic/RegistrationBusinessLogic.cs
qxz2mqe: fixed

src/administration/Administration.Service/BusinessLogic/RegistrationBusinessLogic.cs
qxz2mqe: fixed

src/administration/Administration.Service/BusinessLogic/RegistrationBusinessLogic.cs
```diff
@@ -356,4 +357,43 @@ private static async Task<List<UserRoleData>> GetRoleData(IUserRolesRepository u
 
         return roleData;
     }
+
+    public Task<bool> UpdateCompanyBpn(Guid applicationId, string bpn, string iamuserId)
+    {
+        var regex = new Regex(@"(\w|\d){16}");
+        if (!regex.IsMatch(bpn))
+        {
+            throw new ArgumentException("BPN must contain exactly 16 characters long.", nameof(bpn));
```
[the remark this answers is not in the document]
qxz2mqe: fixed

src/portalbackend/PortalBackend.DBAccess/Repositories/UserRepository.cs
```diff
@@ -494,4 +494,15 @@ public CompanyUser AttachAndModifyCompanyUser(Guid companyUserId, Action<Company
                 companyUser.Company!.CompanyUsers.Any(companyUser => companyUser.IamUser!.UserEntityId == iamUserId),
                 companyUser.CompanyId))
             .SingleOrDefaultAsync();
+    
+    /// <inheritdoc />
+    public IAsyncEnumerable<(bool IsApplicationCompany, string? BusinessPartnerNumber, Guid CompanyId)> GetBpnForIamUserUntrackedAsync(string iamUserId, Guid applicationId, string businessPartnerNumber) =>
+        _dbContext.Companies
+            .AsNoTracking()
+            .Where(company => company.CompanyUsers.Any(companyUser => companyUser.IamUser!.UserEntityId== iamUserId)
```
ntruchsess, line 502: That condition enforces that the logged in user must be a member of the company that the businessPartnerNumber is to be updated. From my understanding this condition is not required (resp. incorrect) as a catena-X administrator shall be able to update the businesspartnernumber of a company that has a pending application. This CatenaX-Admin is certainly not member of any such company.

src/administration/Administration.Service/BusinessLogic/RegistrationBusinessLogic.cs
```diff
@@ -356,4 +357,44 @@ private static async Task<List<UserRoleData>> GetRoleData(IUserRolesRepository u
...
+        if (!bpn.ToUpper().StartsWith("BPNL", StringComparison.OrdinalIgnoreCase))
+        {
+            throw new ControllerArgumentException("businessPartnerNumbers must prefixed with BPNL", nameof(bpn));
+        }
+        return UpdateCompanyBpnAsync(applicationId, bpn, iamuserId);
+    }
+
+    private async Task UpdateCompanyBpnAsync(Guid applicationId, string bpn, string iamuserId)
+    {
+        await foreach (var item in _portalRepositories.GetInstance<IUserRepository>().GetBpnForIamUserUntrackedAsync(iamuserId, applicationId, bpn).ConfigureAwait(false))
```
ntruchsess, line 377: that condition is not suitable here - if no match is found the IAsyncEnumerable will have zero items. Checking a result for being default makes sense when using 'SingleOrDefault()' not AsAsyncEnumerable().

Also processing this result-set while iterating makes it hard to understand the query is mostly for validating preconditions. It's known beforehand only a single dataset is to be updated (and only in case the preconditions are met).

I suggest to do all checks beforehand throwing the appropriate exception if the condition is not met, pick the item that corresponds to the application and then do the update:

like so:

```
var result = await _portalRepositories.GetInstance<IUserRepository>().GetBpnForIamUserUntrackedAsync(iamuserId, applicationId, bpn).ToListAsync().ConfigureAwait(false);
if (!result.Any(item => item.IsApplicationCompany))
{
	throw new NotFoundException($"application {applicationId} not found");
}
if (result.Any(item => !item.IsApplicationCompany))
{
	throw new ConflictException($"BusinessPartnerNumber is already assigned to a different company");
}
var applicationCompanyData = result.Single(item => item.IsApplicationCompany);
if (!applicationCompanyData.IsApplicationPending)
{
	throw new ConflictException($"application {applicationId} for company {applicationCompanyData.CompanyId} is not pending");
}
if (!string.IsNullOrWhiteSpace(applicationCompanyData.BusinessPartnerNumber))
{
	throw new ConflictException($"BusinessPartnerNumber of company {applicationCompanyData.CompanyId} has already been set.");
}
_portalRepositories.GetInstance<ICompanyRepository>().AttachAndModifyCompany(applicationCompanyData.CompanyId, c =>
{
	c.BusinessPartnerNumber = bpn;
});
await _portalRepositories.SaveAsync().ConfigureAwait(false);
```

That code assumes the query is changed to:
```
public IAsyncEnumerable<(bool IsApplicationCompany, bool IsApplicationPending, string? BusinessPartnerNumber, Guid CompanyId)> GetBpnForIamUserUntrackedAsync(string iamUserId, Guid applicationId, string businessPartnerNumber) =>
    _dbContext.Companies
        .AsNoTracking()
        .Where(company => 
            company.CompanyApplications.Any(application => application.Id == applicationId) ||
            company.BusinessPartnerNumber == businessPartnerNumber)
        .Select(company =>new ValueTuple<bool, bool, string?, Guid>(
            company.CompanyApplications.Any(application => application.Id == applicationId),
            company.CompanyStatusId==CompanyStatusId.PENDING,
            company.BusinessPartnerNumber,
            company.Id))
        .AsAsyncEnumerable();
```

ntruchsess (changes requested):

src/administration/Administration.Service/BusinessLogic/RegistrationBusinessLogic.cs
```diff
@@ -356,4 +357,44 @@ private static async Task<List<UserRoleData>> GetRoleData(IUserRolesRepository u
...
     }
+
+    public Task UpdateCompanyBpn(Guid applicationId, string bpn, string iamuserId)
+    {
+        var regex = new Regex(@"(\w|\d){16}");
+        if (!regex.IsMatch(bpn))
+        {
+            throw new ControllerArgumentException("BPN must contain exactly 16 characters long.", nameof(bpn));
+        }
+        if (!bpn.ToUpper().StartsWith("BPNL", StringComparison.OrdinalIgnoreCase))
```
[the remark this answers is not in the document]
qxz2mqe: fixed

src/portalbackend/PortalBackend.DBAccess/Repositories/UserRepository.cs
qxz2mqe: fixed

src/administration/Administration.Service/BusinessLogic/RegistrationBusinessLogic.cs
qxz2mqe: implemented the change

ntruchsess (approved):

PR #344, opened 2022-11-22
CPLP-1772: Adjust appsettings

evegufy:
Add missing objects, present in secrets but not in appsettings.
Remove env specific urls and change to test urls.
Noticed in the course of https://jira.catena-x.net/browse/CPLP-1772


ntruchsess (changes requested):
please use urls that are obviously 'dummy'-urls (like (https://null) and add the real urls to the secrets.

evegufy:
> please use urls that are obviously 'dummy'-urls (like (https://null) and add the real urls to the secrets.

done

PR #346, opened 2022-11-23
CPLP-1601-added-migration-for-notificationenum

qxz2mqe:
https://jira.catena-x.net/browse/CPLP-1601


ntruchsess (changes requested):
as the Migration that has been implemented with CPLP-1318 is already approved (see https://github.com/catenax-ng/product-portal-backend/pull/335 ) this new migration for CPLP-1601 shall build upon it. We must not create migrations in parallel merging them in git.

ntruchsess (approved):

PR #347, opened 2022-11-23
bugfix/TEST-174-invite-email-shows-uuid

qxz2mqe:
https://jira.catena-x.net/browse/TEST-174


ntruchsess (approved):
I did refactor removing some redundancy re-using existing code of UserProvisioningService. That code takes into account that FirstName, LastName and Email are nullable in the database and does best effort to create a reasonable Name-string out of those fields: https://github.com/catenax-ng/product-portal-backend/blob/630c040d0a338ad7c4a534fc3d185775ded0a514/src/provisioning/Provisioning.Library/Service/UserProvisioningService.cs#L202

PR #350, opened 2022-11-24
CPLP-1766 provider can be updated

Phil91:


ntruchsess (approved):
despite my comment I did approve this PR. If you feel my comment is worth to implement feel free to do so :-)
I did do some cleanup in the unit-tests renaming Notification in the namespace to Notifications (to avoid the clash on class vs. namespace) Please double-check before merging

PR #352, opened 2022-11-25
Email Templates Styling

jjeroch:


src/mailing/Mailing.Template/EmailTemplates/serviceprovider_subscription_request.html
```diff
@@ -1,219 +1,177 @@
...
-    <meta name="viewport" content="width=device-width">
-    <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
-    <title>Register to Catena-X Welcome Email</title>
-    <!--[if mso 14]>
+    <meta charset="UTF-8">
+    <meta content="width=device-width, initial-scale=1" name="viewport">
+    <meta name="x-apple-disable-message-reformatting">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta content="telephone=no" name="format-detection">
+    <title>Invite</title><!--[if (mso 16)]>
```
Phil91, line 9: The title did change to invite, that should not make any problems, but maybe we want to change it to _Catena-X Service Subscription Request_ for consistency reasons?

src/mailing/Mailing.Template/EmailTemplates/serviceprovider_subscription_activation.html
```diff
@@ -1,213 +1,130 @@
...
-    <meta name="viewport" content="width=device-width">
-    <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
-    <title>Catena-X Service Subscription Activation</title>
-    <!--[if mso 14]>
+    <meta charset="UTF-8">
+    <meta content="width=device-width, initial-scale=1" name="viewport">
+    <meta name="x-apple-disable-message-reformatting">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta content="telephone=no" name="format-detection">
+    <title>Invite</title><!--[if (mso 16)]>
```
Phil91, line 9: The title did change as mentioned below, should not make any problems nor a user would see this in his mail program, but we may want to change that back?

Phil91 (commented):
Other than the two title comments, the emails look fine

jjeroch:
@Phil91 quick question regarding the app subscription. Are you using the service subscription activation template for apps as well or do we do not have any "App Subscription Activation" message/email so far?

Phil91:
> @Phil91 quick question regarding the app subscription. Are you using the service subscription activation template for apps as well or do we do not have any "App Subscription Activation" message/email so far?

@jjeroch currently we just create a notification. there is no mail which gets send. I could add a mail which gets send when activating the app subscription. shouldn't be a lot of effort

Phil91:
> > @Phil91 quick question regarding the app subscription. Are you using the service subscription activation template for apps as well or do we do not have any "App Subscription Activation" message/email so far?
> 
> @jjeroch currently we just create a notification. there is no mail which gets send. I could add a mail which gets send when activating the app subscription. shouldn't be a lot of effort

@jjeroch i've added the logic to send the app activation mail

Phil91 (approved):

Phil91:
@ntruchsess could you please review this PR as well since i added the functionallity to send the subscriptionActivation  Mail

src/marketplace/Apps.Service/BusinessLogic/AppsBusinessLogic.cs
```diff
@@ -195,6 +199,16 @@ public async Task ActivateOwnCompanyProvidedAppSubscriptionAsync(Guid appId, Gui
                 });
             });
         
+        if (!string.IsNullOrWhiteSpace(email))
+        {
+            var mailParams = new Dictionary<string, string>
+            {
+                { "offerCustomerName", firstname ?? "User" },
+                { "offerName", appName! },
```
ntruchsess, line 207: either we validate and throw ConflictException in case AppName is not yet set or we provide a default like 'appName ?? "unknown App"'.

src/portalbackend/PortalBackend.DBAccess/Repositories/OfferSubscriptionsRepository.cs
```diff
@@ -84,14 +84,16 @@ sorting switch
...
             .Where(app => app.Id == appId)
-            .Select(app => new ValueTuple<OfferSubscription?, bool, string?, Guid>(
+            .Select(app => new ValueTuple<OfferSubscription?, bool, string?, Guid, string?, string?>(
                 app.OfferSubscriptions.SingleOrDefault(assignedApp => assignedApp.CompanyId == companyId),
                 app.ProviderCompany!.CompanyUsers.Any(companyUser => companyUser.IamUser!.UserEntityId == iamUserId),
                 app.Name,
-                app.ProviderCompany!.CompanyUsers.Any(companyUser => companyUser.IamUser!.UserEntityId == iamUserId) ? app.ProviderCompany!.CompanyUsers.First(companyUser => companyUser.IamUser!.UserEntityId == iamUserId).Id : Guid.Empty
+                app.ProviderCompany!.CompanyUsers.Any(companyUser => companyUser.IamUser!.UserEntityId == iamUserId) ? app.ProviderCompany!.CompanyUsers.First(companyUser => companyUser.IamUser!.UserEntityId == iamUserId).Id : Guid.Empty,
+                app.OfferSubscriptions.SingleOrDefault(assignedApp => assignedApp.CompanyId == companyId) == null ? app.OfferSubscriptions.Single(assignedApp => assignedApp.CompanyId == companyId).Requester!.Email : null,
```
ntruchsess, line 95: that's will not improve anything as this will translate into sql that is 100% redundant. In SQL you just join everything that needs to be joined and the database then traverses the joined tables as it finds data and stops traversing if there is none.
Also isMemberOfCompanyProvidingApp is redundant with companyUserId (being empty in case the user is not found).
I suggest:
```
public Task<(OfferSubscription? companyAssignedApp, string? appName, Guid companyUserId, string? email, string? firstname)> GetCompanyAssignedAppDataForProvidingCompanyUserAsync(Guid appId, Guid companyId, string iamUserId) =>
    _context.Offers
        .Where(app => app.Id == appId)
        .Select(app => new {
            App = app,
            OfferSubscription = app.OfferSubscriptions.SingleOrDefault(assignedApp => assignedApp.CompanyId == companyId),
        })
        .Select(x => new ValueTuple<OfferSubscription?, string?, Guid, string?, string?>(
            x.OfferSubscription,
            x.App.Name,
            x.App.ProviderCompany!.CompanyUsers.SingleOrDefault(companyUser => companyUser.IamUser!.UserEntityId == iamUserId)!.Id,
            x.OfferSubscription!.Requester!.Email,
            x.OfferSubscription.Requester.Firstname
        ))
        .SingleOrDefaultAsync();
```

ntruchsess (changes requested):

src/marketplace/Apps.Service/BusinessLogic/AppsBusinessLogic.cs
Phil91: At this point the app should have the name set, thus we go with the ConflictException

src/portalbackend/PortalBackend.DBAccess/Repositories/OfferSubscriptionsRepository.cs
Phil91: edited

ntruchsess (approved):

PR #354, opened 2022-11-29
CPLP-1518 approve app release

qxz2mqe:
https://jira.catena-x.net/browse/CPLP-1518


src/portalbackend/PortalBackend.DBAccess/Repositories/OfferRepository.cs
```diff
@@ -462,4 +462,17 @@ sorting switch
                 offer.ProviderCompany!.CompanyUsers.SingleOrDefault(companyUser => companyUser.IamUser!.UserEntityId == userId)!.Id
             ))
             .SingleOrDefaultAsync();
+
+    ///<inheritdoc/>
+    public Task<(bool OfferExists, bool IsStatusInReview, bool IsProviderCompanyUser, string OfferName)> GetOfferStatusDataByIdAsync(Guid appId, string iamUserId, OfferTypeId offerTypeId) =>
+        _context.Offers
+            .Where(offer => offer.Id == appId && offer.OfferTypeId == offerTypeId)
+            .Select(offer => new ValueTuple<bool, bool, bool, string>(
+                true,
```
Phil91, line 471: This bool value is not needed. You can check in the BL if the result is == default. When no offer exists for the given id the default would be returned.

src/marketplace/Apps.Service/BusinessLogic/AppReleaseBusinessLogic.cs
```diff
@@ -532,4 +532,39 @@ public async Task SubmitAppReleaseRequestAsync(Guid appId, string iamUserId)
         await _notificationService.CreateNotifications(_settings.CompanyAdminRoles, requesterId, content).ConfigureAwait(false);
         await _portalRepositories.SaveAsync().ConfigureAwait(false);
     }
+    
+    /// <inheritdoc/>
+    public async Task ApproveAppRequestAsync(Guid appId, string iamUserId)
+    {
+        var offerRepository = _portalRepositories.GetInstance<IOfferRepository>();
+        var appDetails = await offerRepository.GetOfferStatusDataByIdAsync(appId, iamUserId, OfferTypeId.APP).ConfigureAwait(false);
+        if (!appDetails.OfferExists)
```
Phil91, line 541: this check must be changed to `appDetails == default` since the repository would return a default value when the given app does not exists

Phil91 (changes requested):
Please add unit tests as well

src/portalbackend/PortalBackend.DBAccess/Repositories/OfferRepository.cs
qxz2mqe: fixed

src/marketplace/Apps.Service/BusinessLogic/AppReleaseBusinessLogic.cs
qxz2mqe: fixed

qxz2mqe:
@Phil91 I have added the Unit Test

Phil91:
@qxz2mqe the implementation looks good for me. But before approving the PR I'd like to discuss with you and @jjeroch if we really don't want to add a user validation at all. With this implementation everyone with the role _approve_app_release_ would be able to approve the app.

Phil91:
> @qxz2mqe the implementation looks good for me. But before approving the PR I'd like to discuss with you and @jjeroch if we really don't want to add a user validation at all. With this implementation everyone with the role _approve_app_release_ would be able to approve the app.

After discussing we will have a followup story in the future to add a check if the current user is assigned to an operating company.

src/marketplace/Apps.Service/BusinessLogic/AppReleaseBusinessLogic.cs
```diff
@@ -537,4 +537,36 @@ public async Task SubmitAppReleaseRequestAsync(Guid appId, string iamUserId)
...
+    {
+        var offerRepository = _portalRepositories.GetInstance<IOfferRepository>();
+        var appDetails = await offerRepository.GetOfferStatusDataByIdAsync(appId, OfferTypeId.APP).ConfigureAwait(false);
+        if (appDetails == default)
+        {
+            throw new NotFoundException($"Apps not found. Either Not Existing or incorrect offer type");
+        }
+        if (!appDetails.IsStatusInReview)
+        {
+            throw new ConflictException($"Apps in InCorrect Status {appDetails.IsStatusInReview} ");
```
Phil91, line 552: Just recognized, the message would always be _Apps in InCorrect Status false_ if we want to get the current status of the app the repository should return the status instead of the bool identifier if the app is in the correct state

src/portalbackend/PortalBackend.DBAccess/Repositories/OfferRepository.cs
```diff
@@ -461,4 +461,15 @@ sorting switch
...
             ))
             .SingleOrDefaultAsync();
+
+    ///<inheritdoc/>
+    public Task<(bool IsStatusInReview, string OfferName)> GetOfferStatusDataByIdAsync(Guid appId, OfferTypeId offerTypeId) =>
+        _context.Offers
+            .Where(offer => offer.Id == appId && offer.OfferTypeId == offerTypeId)
+            .Select(offer => new ValueTuple<bool, string>(
+                offer.OfferStatusId == OfferStatusId.IN_REVIEW,
+                offer.Name!
```
Phil91, line 471: We should return the nullable value here and throw a ConflictException in the BusinessLogic if the AppName is null

Phil91 (changes requested):

src/marketplace/Apps.Service/BusinessLogic/AppReleaseBusinessLogic.cs
qxz2mqe: fixed

src/portalbackend/PortalBackend.DBAccess/Repositories/OfferRepository.cs
qxz2mqe: fixed

Phil91 (approved):

PR #358, opened 2022-11-30
CPLP-1833 company role consent

Phil91:
In this PR [CPLP-1833](https://jira.catena-x.net/browse/CPLP-1833) is resolved.


src/registration/Registration.Service/BusinessLogic/RegistrationBusinessLogic.cs
```diff
@@ -517,6 +501,42 @@ public async Task<RegistrationData> GetRegistrationDataAsync(Guid applicationId,
...
 
+    private static void HandleConsent(IEnumerable<Consent> consents, IEnumerable<AgreementConsentStatus> agreementConsentsToSet,
+        IConsentRepository consentRepository, Guid companyId, Guid companyUserId)
+    {
+        var consentsToInactivate = consents
+            .Where(consent =>
+                !agreementConsentsToSet.Any(agreementConsent =>
+                    agreementConsent.AgreementId == consent.AgreementId
+                    && agreementConsent.ConsentStatusId == ConsentStatusId.ACTIVE));
+        consentRepository.AttachAndModifiesConsents(consentsToInactivate, consent =>
```
ntruchsess, line 512: attaching tracked entities is pointless:
https://github.com/catenax-ng/product-portal-backend/blob/1936198c234559b1559b73524b6a2873fe07201b/src/portalbackend/PortalBackend.DBAccess/Repositories/CompanyRolesRepository.cs#L49

(We should consider to change the query to return only those fields that are required to identify and apply the changes instead of full tracked entities).

ntruchsess (changes requested):

src/registration/Registration.Service/BusinessLogic/RegistrationBusinessLogic.cs
Phil91: did a bit of refactoring.

ntruchsess (approved):
I did equivalent refactoring in respect to applicationAssingedRoles adjusting the unit-tests accordingly. Please double-check before merging

PR #359, opened 2022-12-01
CPLP-1850 edit search for app and service

Phil91:


src/portalbackend/PortalBackend.DBAccess/Repositories/OfferSubscriptionsRepository.cs
```diff
@@ -64,7 +64,7 @@ public OfferSubscriptionsRepository(PortalDbContext portalDbContext)
                     .Where(os => 
                         os.OfferTypeId == offerTypeId &&
                         os.ProviderCompany!.CompanyUsers.Any(companyUser => companyUser.IamUser!.UserEntityId == iamUserId) &&
-                        (statusId == null || os.OfferSubscriptions.Any(x => x.OfferSubscriptionStatusId == statusId)))
+                        os.OfferSubscriptions.Any(x => x.OfferSubscriptionStatusId == statusId))
```
ntruchsess, line 67: that change alone is not sufficient. It filters the offers to include those that have at lease 1 active subscription. To fulfill https://jira.catena-x.net/browse/CPLP-1850 requirements the subscriptions that are included in the result must be filtered as well.
line 81 needs to be changed to:
```
CompanySubscriptionStatuses = g.OfferSubscriptions
    .Where(os => os.OfferSubscriptionStatusId == statusId)
    .Select(s => new CompanySubscriptionStatusData(s.CompanyId, s.Company!.Name, s.Id, s.OfferSubscriptionStatusId))
```

ntruchsess (changes requested):

src/portalbackend/PortalBackend.DBAccess/Repositories/OfferSubscriptionsRepository.cs
Phil91: filter and unit test added

ntruchsess (approved):

PR #366, opened 2022-12-06
Add editorconfig and dotnet format check

Phil91:
Only a test for dotnet format


Phil91:
PR was only for testing purposes. Will be completly implemented after the repo migration

PR #367, opened 2022-12-06
CPLP-1828-registration-document-refactor

qxz2mqe:
https://jira.catena-x.net/browse/CPLP-1828

### DIFF
--- a/src/portalbackend/PortalBackend.DBAccess/Repositories/DocumentRepository.cs
+++ b/src/portalbackend/PortalBackend.DBAccess/Repositories/DocumentRepository.cs
@@ -68,12 +68,10 @@ public class DocumentRepository : IDocumentRepository
                     document.CompanyUser!.IamUser!.UserEntityId == iamUserId))
             .SingleOrDefaultAsync();
 
-    public IAsyncEnumerable<UploadDocuments> GetUploadedDocumentsAsync(Guid applicationId, DocumentTypeId documentTypeId, string iamUserId) =>
+    public IAsyncEnumerable<UploadDocuments> GetUploadedDocumentsAsync(Guid applicationId, DocumentTypeId documentTypeId) =>
         _dbContext.IamUsers
             .AsNoTracking()
-            .Where(iamUser =>
-                iamUser.UserEntityId == iamUserId
-                && iamUser.CompanyUser!.Company!.CompanyApplications.Any(application => application.Id == applicationId))
+            .Where(iamUser => iamUser.CompanyUser!.Company!.CompanyApplications.Any(application => application.Id == applicationId))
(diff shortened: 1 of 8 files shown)


src/portalbackend/PortalBackend.DBAccess/Repositories/DocumentRepository.cs
```diff
@@ -68,12 +68,10 @@ public Document CreateDocument(string documentName, byte[] documentContent, byte
...
             .SingleOrDefaultAsync();
 
-    public IAsyncEnumerable<UploadDocuments> GetUploadedDocumentsAsync(Guid applicationId, DocumentTypeId documentTypeId, string iamUserId) =>
+    public IAsyncEnumerable<UploadDocuments> GetUploadedDocumentsAsync(Guid applicationId, DocumentTypeId documentTypeId) =>
         _dbContext.IamUsers
             .AsNoTracking()
-            .Where(iamUser =>
-                iamUser.UserEntityId == iamUserId
-                && iamUser.CompanyUser!.Company!.CompanyApplications.Any(application => application.Id == applicationId))
+            .Where(iamUser => iamUser.CompanyUser!.Company!.CompanyApplications.Any(application => application.Id == applicationId))
```
Phil91, line 74: we should add a check if the user is part of the company of the given application, otherwise any user could see the documents for a specific application

Phil91 (changes requested):

src/portalbackend/PortalBackend.DBAccess/Repositories/DocumentRepository.cs
qxz2mqe: As part of this requirement the documents uploaded by another user under the same application id should also be visible . Earlier we had validation where executing user can see only document uploaded by himself

Phil91: correct, but we still should check that the users are in the same company of the given application, if we don't a user of company a would be able to access a document of company b with the application id, that would be a huge security issue.

Phil91: oversaw the where check - all good.

Phil91 (approved):

src/portalbackend/PortalBackend.DBAccess/Repositories/DocumentRepository.cs
ntruchsess: actually the query does not check the calling user is associated with the application. It will return documents of any application regardless whether the company of the calling user is the one that applies.
(Side note: having removed any iam-user-related criteria it's nonsense to start the query at 'IamUsers' given that none of the fields is being referenced any more. If the criteria would be correct the query then should rather start at 'Documents')

ntruchsess (changes requested):

